PR #8, opened 2017-01-03
Miscellaneous smaller impovements

sschuberth:
Please have a look at the individual commit messages for the details.


pombredanne:
Good catch .. The semantics of `and` are correct and using `&` was not at all...

pombredanne:
Merged in http://git.spdx.org/spdx-tools-python.git/ .... and it will eventually mirror here.

Issue #4, opened 2016-11-25
Setup CI on travis, appveyor and circle


pombredanne:

pombredanne:
As a followup to @sschuberth in #14 we will also enable Circle

sschuberth:
Fixed via https://github.com/spdx/tools-python/pull/18.

pombredanne:
@sschuberth You rock. Let me create another ticket as we would want the status badges in the readme for these!

PR #12, opened 2017-01-16
Allow to create a file with pre-calculated SHA1

sschuberth:
And use a file's SHA1, if present, in package verification code
calculation.

### DIFF
--- a/spdx/package.py
+++ b/spdx/package.py
@@ -192,7 +192,12 @@ class Package(object):
         hashes = []
 
         for file_entry in self.files:
-            hashes.append(file_entry.calc_chksum())
+            if isinstance(file_entry.chk_sum, checksum.Algorithm) and
+               file_entry.chk_sum.identifier == 'SHA1':
+                sha1 = file_entry.chk_sum.value
+            else:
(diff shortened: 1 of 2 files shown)


spdx/package.py
```diff
@@ -192,7 +192,12 @@ def calc_verif_code(self):
         hashes = []
 
         for file_entry in self.files:
-            hashes.append(file_entry.calc_chksum())
+            if isinstance(file_entry.chk_sum, checksum.Algorithm) and
```
pombredanne: Note that this is not a valid syntax and the multline if should be wrapped in parens. I fixed this and pushed that to the LF git repo.

pombredanne:
Merged and pushed to http://git.spdx.org/spdx-tools-python.git/ Thanks!

pombredanne:
and this has been pushed to Pypi

pombredanne:
as v0.4.1

spdx/package.py
sschuberth: Thanks. I probably need to work on #4 next so I get slapped for blindly typed stuff like that :-D

Issue #10, opened 2017-01-06
Add full support for license expressions


pombredanne:
https://github.com/nexB/license-expression provides a comprehensive support for parsing, comparing, validating, normalizing and resolving license expressions based on SPDX or any other license identifiers or names. It uses a boolean logic engine behind the scenes to handle this correctly

We should add this here to support SPDX license expressions here.

yash-nisar:
I would like to work on this @pombredanne :)

aviral1701:
@yash-nisar this is a GSoC project.I don’t think you should work on this now.

yash-nisar:
Oh, I didn't notice that this is a GSoC project. However, we should let @pombredanne decide this. :)

thunda99:
@pombredanne 
Hi, I would like to work on this for GSoC2018. I so love this project. I have started by having a close look at https://github.com/nexB/license-expression, to better understand this project and requirements.

Thanks

thunda99:
@pombredanne 
After looking and running tests for tools-python, 
I noticed that the code structure is good. 
The use of ply, a good choice.
I have also looked at [nexB](https://github.com/nexB/license-expression).

My question is, how do you want this project to go?
Do you want want this project to be an integration with [nexB](https://github.com/nexB/license-expression) or you want us to be inspired by it and have our codebase duplicate some of his logic?

bhavybarca:
@pombredanne i would like to work on the issue, in what direction should i explore to make the first contribution i.e. what should be my first deliverable to this issue. 
Thanks

pombredanne:
@bhavybarca You would need to design how to integrate support for this library https://github.com/nexB/license-expression/  and that would also require updating all the internal models accordingly. This is a project of its own alright.

veer11997:
Hi, I would like to work on this for GSoC 2019. I would like to work on the project.

nicoweidner:
See https://github.com/spdx/tools-python/pull/111 for an implementation attempt. I don't think continuing with that PR is viable though, as it is very old and outdated.

armintaenzertng:
The `license_expression` library is now integrated into the refactored model. As this is not merged to main yet, I will leave this issue open.

armintaenzertng:
The refactored version can now be found on main, so I'll go ahead and close this issue. Please speak up if it should be reopened.

PR #13, opened 2017-02-08
Improvements upstream

hugo-bmwcarit:


pombredanne:
@hugo-bmwcarit Thanks ++ for all this! Let me review this in details tomorrow

pombredanne:
@hugo-bmwcarit @neverpanic As a side note since I see you cleaned up some license conjunction/disjunction handling code,  see also #10 . I have a library that will integrate soon to handle properly flexible expressions parsing and a comprehensive boolean logic on these.

spdx/document.py
```diff
@@ -17,6 +17,15 @@
 from spdx.version import Version
 
 
+def _add_parens(required, text):
+    """
+    Add parens around a license expression if `required` is True, otherwise
+    return `text` unmodified.
+    """
+    if required:
+        return '(' + str(text) + ')'
```
pombredanne: I wonder if using interpolation or format would not be nicer?
e.g. either:
- `return '(%s)' % (text)`
- `return '({text})'.format(a=text)`
and then maybe this simpler form (though the if construct is fine too):
- `return required and '({text})'.format(a=text) or text`

spdx/parsers/tagvaluebuilders.py
```diff
@@ -36,7 +36,7 @@ def checksum_from_sha1(value):
     checksum. Returns None if doesn't match CHECKSUM_RE.
     """
     # More constrained regex at lexer level
-    CHECKSUM_RE = re.compile('SHA1:(.+)', re.UNICODE)
+    CHECKSUM_RE = re.compile('SHA1: *(.+)', re.UNICODE)
```
pombredanne: Are tabs or other whitespaces also OK there? e.g. would `CHECKSUM_RE = re.compile('SHA1:\s*(.+)', re.UNICODE)` be better or may be it does not make sense? 

And would it make sense to drop trailing or inner whitespaces entirely e.g. may be with `CHECKSUM_RE = re.compile('SHA1:\s*([\S]+)', re.UNICODE)` ?

spdx/utils.py
```diff
@@ -114,7 +114,7 @@ def t_whitespace(self, t):
         pass
 
     def t_LICENSE(self, t):
-        r'[A-Za-z.0-9\-]+'
+        r'[A-Za-z.0-9\-+]+'
```
pombredanne: Good catch!

spdx/writers/rdf.py
```diff
@@ -151,9 +152,9 @@ def create_license_node(self, lic):
         Could be a single license (extracted or part of license list.) or
         a conjunction/disjunction of licenses.
         """
-        if isinstance(lic, document.LicenseConjuction):
+        if isinstance(lic, document.LicenseConjunction):
             return self.create_conjunction_node(lic)
-        elif isinstance(license, document.LicenseDisjunction):
+        elif isinstance(lic, document.LicenseDisjunction):
```
pombredanne: Nice find!

stdeb.cfg
```diff
@@ -0,0 +1,4 @@
+[DEFAULT]
+Depends = python-rdflib (>= 4.1.2), python-ply (>= 3.4), python-nose (>= 1.3)
+Copyright-File = LICENSE
```
pombredanne: Ah! great... I guess we will need a machine-readable copyright file in the future then.... But doss `stddeb` require its config file to be at the root? or should we move it to a `debian` sub-directory?

Also, would it be possible to have a minimal smoke test that would validate a package creation? 

NB: this reminds me that I dislike having a hard dependency on `nose` for tests. I created #15 tp track this.

spdx/document.py
neverpanic: I'd use
```python
return '({})'.format(text) if required else text
```
then.

spdx/parsers/tagvaluebuilders.py
neverpanic: I agree for both.

pombredanne:
@hugo-bmwcarit @neverpanic I merged this and added the few fixes from the review. Thank you!

neverpanic:
@pombredanne Thanks. Sorry it fell off my radar.

hugo-bmwcarit:
@pombredanne, thanks for merging that. I'm still testing your changes for that other project and didn't finish the review yet. For the SPDX Python tools, thanks for merging that and doing the changes by yourself.

hugo-bmwcarit:
@neverpanic, I was supposed to take care of that, but @pombredanne did it faster than me.

pombredanne:
@hugo-bmwcarit These were really minor fixes, do not mention it :)

PR #14, opened 2017-02-10
Add basic Travis CI integration

sschuberth:
This is to have PRs checked at least on *one* platform. Partly addresses issue #4.

@pombredanne Since I was having problems getting this to run on OS X (see the commit message of 79c220f), and also because the OS X nodes in Travis are very slow / builds spend a long time in the queue, I was thinking about using CircleCI for OS X builds. If then Windows builds would run on AppVeyor, we'd use a dedicated service per platform, probably speeding up the checks as all of them are more likely to run in parallel. What do you think?


pombredanne:
This is a great idea and makes perfect sense. Let me sign up for Travis and Appveyor for now and I will sign up for Circle too.

NB: it is rare to make `setup.py` an executable file: any reason why you did so?

sschuberth:
> NB: it is rare to make `setup.py` an executable file: any reason why you did so?

Well, I was getting [./setup.py: Permission denied](https://travis-ci.org/sschuberth/spdx-tools-python/jobs/200011649), which is correct, as it is tried to be executed directly (with `PYTHON_EXE` being set). Also, I noticed `setup.py` has a shebang line, so I though it *should* be executable.

Would you prefer to not make it executable, and call whatever `PYTHON_EXE` is set to directly and have it execute `setup.py`?

pombredanne:
@sschuberth good points. Just curious!

pombredanne:
Merged. Thanks!

sschuberth:
> I will sign up for Circle too

BTW, CircleCI can either build on Linux (the default), *or* on OS X, but not both in the same (matrix) build. Also, for Open Source projects to get access to OS X nodes, you have to [contact billing](https://github.com/circleci/frontend/blob/master/src-cljs/frontend/components/org_settings.cljs#L286). But I guess it's worth the effort.

pombredanne:
:+1:  For travis you need to contact support for OSX not billing ;)

sschuberth:
Hmm, looking at the Travis build logs again, they don't seem to build for the defined multiple Python version, only for Python 2.7. Looks like you have the same issue in the license-expression project. Looks like the `python:` key does not work for the `cpp` language. Or am I reading the log wrong?

PR #16, opened 2017-02-10
More CI fixes, incl. Python 3 support

sschuberth:
@pombredanne This fixes Travis CI matrix builds for different Python version, and adds AppVeyor CI builds.

Note that the builds will actually fail, because this now properly builds with Python 3 for the first time, and there seem to be [some](https://travis-ci.org/sschuberth/spdx-tools-python/jobs/200283107) incompatibility [issues](https://ci.appveyor.com/project/sschuberth/spdx-tools-python/build/10/job/qwe7sf0ynaajm4g5) in the source code that need to get fixed. I'll add these fixes to this PR while discussing it.


sschuberth:
@pombredanne The commit 4b1355c finalizes Python 3 support.

sschuberth:
And the CircleCI build on macOS works now, too.

.gitignore
```diff
@@ -5,3 +5,4 @@ __pycache__/
 /build/
 /dist/
 /tmp/
+spdx/parsers/parsetab.py
```
pombredanne: What was the rationale to remove this file entirely? I know it is generated, but was removing it because of Python3 changes?

pombredanne: BTW, this is likely related. https://github.com/dabeaz/ply/issues/31

sschuberth: No, this change is not related to Python-3-porting. It just annoyed me that the file was being regenerated all the time on my local Python-2-based environment. Do, would you suggest to upgrade PLY instead? Or is this change still a good one as we should not commit generated files in the first place?

sschuberth:
@pombredanne Thanks for merging. Have you enabled Travis CI / AppVeyor / CircleCI integration for this repo yet? On a somewhat related note, I was considering asking you and @goneall to become part of the SPDX org, so I could do stuff like this myself. What do you think?

pombredanne:
@sschuberth I started added circle alright, but if you can help on the other that would be great.

.gitignore
pombredanne: ok, so this is a problem with Ply. I cannot recall why I started  committing generated file but there was an issue AFAIRC, may be with wheels.

pombredanne:
@sschuberth as long as as you agree to the https://developercertificate.org/ ... we are good.
@goneall unless you have an objection, I will make @sschuberth a full admin of this repo, such that he can setup the CI services correctly.

sschuberth:
Yes, I do accept https://developercertificate.org/, and will send you a mail with more details.

goneall:
@pombredanne No objections to making @sschuberth a full admin of this repo.

PR #18, opened 2017-02-27
Fix CircleCI integration, and test all CI

sschuberth:
The default machine step calls virtualenv, which is not installed. Doh!


pombredanne:
May be this had made you appreciate the `configure` script in scancode a bit more :)

PR #25, opened 2017-03-31
Add transparent pytest support

all3fox:
Before:

```
./configure
source ./bin/activate
pip install pytest pytest-cov pytest-xdist
pytest --cov=spdx tests
```

Results in `spdx` not found, then `test_utils` not found

After:

The command sequence above runs fine and reports coverage

Signed-off-by: Aleksandr Lisianoi <all3fox@gmail.com>

### DIFF
--- a/tests/test_builder.py
+++ b/tests/test_builder.py
@@ -16,7 +16,7 @@ from __future__ import print_function
 import unittest
 from unittest import TestCase
 
-import testing_utils
+import tests.testing_utils as testing_utils
 
 from spdx.document import Document, License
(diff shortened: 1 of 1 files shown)


tests/test_builder.py
```diff
@@ -16,7 +16,7 @@
 import unittest
 from unittest import TestCase
 
-import testing_utils
+import tests.testing_utils as testing_utils
```
pombredanne: Good catch. We should use the future absolute_import everywhere quite likely. As part of porting to Python 3 sometimes.

PR #35, opened 2018-02-28
Travis Build Fixed

yugeshk:
I'm not sure if this is the right fix - or if this is a proper fix at all, but from what I saw in the Travis logs, I guess this does the trick.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,7 +1,10 @@
 machine:
+  ruby:
+    version: 2.3.3
   pre:
     - python --version
     - sudo -H pip install --upgrade virtualenv
+    - brew update
     - brew install pyenv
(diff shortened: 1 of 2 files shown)


yugeshk:
@pombredanne This patch fixes the other builds as well, not just Travis.

sschuberth:
I just realized https://github.com/spdx/tools-python/pull/37 does almost the same. I tend to close this PR in favor of the other, as the other PR splits unrelated changes into different commits.

yugeshk:
@sschuberth - will keep that in mind from next time. Sorry :sweat_smile:

yugeshk:
However, I could make amend the PR to split it down, if thats the case.

Issue #36, opened 2018-03-04
Build failure for Travis, Appveyor and Circle


yash-nisar:
[Travis](https://travis-ci.org/yash-nisar/tools-python/jobs/348863131) and [Appveyor](https://ci.appveyor.com/project/yash-nisar/tools-python/build/job/6ip42wycsy5wtna9) builds seem to be failing.

Also, it would be great to add support for Python 3.6.

yash-nisar:
I'm working on a quick fix to get the builds green again.

yash-nisar:
After a few failed attempts, I finally managed to get the builds passing on [Travis](https://travis-ci.org/yash-nisar/tools-python/builds/348877226) and [Appveyor](https://ci.appveyor.com/project/yash-nisar/tools-python/build/4).

From the [docs](https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files):
> 'b' appended to the mode opens the file in binary mode: now the data is read and written in the form of bytes objects. This mode should be used for all files that don’t contain text.

For our CI systems running on Python3: 
1. Literal strings are unicode by default in Python3.
2. Python3 returns 'bytes' and not 'str' when reading/writing to/from a binary stream. So, the files need to be opened in text mode for `str` related operations.

PR #37, opened 2018-03-04
Fix build failure for Travis and Appveyor #36

yash-nisar:
Python3 returns 'bytes' and not 'str' when reading/writing
to/from a binary stream. So, the files need to be opened in
text mode for `str` related operations.

Signed-off-by: Yash Nisar <yash.nisar@somaiya.edu>

Closes https://github.com/spdx/tools-python/issues/36

### DIFF
--- a/tests/test_document.py
+++ b/tests/test_document.py
@@ -188,7 +188,7 @@ class TestWriters(TestCase):
         try:
             temp_dir = tempfile.mkdtemp(prefix='test_spdx')
             result_file = os.path.join(temp_dir, 'spdx-simple.tv')
-            with open(result_file, 'wb') as output:
+            with open(result_file, 'w') as output:
                 write_document(doc, output, validate=True)
 
             expected_file = utils_test.get_test_loc(
@@ -211,7 +211,7 @@ class TestWriters(TestCase):
             result_file = os.path.join(temp_dir, 'spdx-simple-plus.tv')
 
             # test proper!
-            with open(result_file, 'wb') as output:
+            with open(result_file, 'w') as output:
                 write_document(doc, output, validate=True)
 
             expected_file = utils_test.get_test_loc(


yash-nisar:
@pombredanne A quick review possible ?

yash-nisar:
Circle CI states that: 
```
/usr/local/Homebrew/Library/Homebrew/brew.rb:12:in `<main>': Homebrew must be run under Ruby 2.3! You're running 2.0.0. (RuntimeError)
```

Upgrading Ruby to 2.3 should solve the problem.

yash-nisar:
@a-h-i @sschuberth  Would you like to have a look ?

circle.yml
```diff
@@ -1,5 +1,6 @@
 machine:
   pre:
+    - brew update
```
sschuberth: Please explain in the commit message why this is necessary, because AFAIK `brew install` does an auto-update.

tests/test_document.py
```diff
@@ -188,7 +188,7 @@ def test_write_document_tv_with_validate(self):
         try:
             temp_dir = tempfile.mkdtemp(prefix='test_spdx')
             result_file = os.path.join(temp_dir, 'spdx-simple.tv')
-            with open(result_file, 'wb') as output:
+            with open(result_file, 'w') as output:
```
sschuberth: Sorry, I'm not a Python expert: Will this work for Python 2, too? Because we need to stay compatible to both versions 2 and 3.

yash-nisar: Yes, this will work on Python 2 too. The tests pass for Python 2.7. :)

sschuberth: Then I'm wondering why it ever worked for Python 3. Do you have any idea about that?

yash-nisar: Yes, Python 3 returns 'bytes' and not 'str' when reading/writing to/from a binary stream. So, the files need to be opened in text mode for `str` related operations.

circle.yml
yash-nisar: Some useful threads which I found were -> https://discuss.circleci.com/t/homebrew-must-be-run-under-ruby-2-3-runtimeerror/17232 and https://github.com/Homebrew/brew/issues/3299#issuecomment-335837083

tests/test_document.py
yugeshk: @sschuberth I guess the answer is in the Travis Builds - If you look at older builds that [passed](https://travis-ci.org/spdx/tools-python/jobs/243330760) the tests hadn't been added up till then (back when the builds were passing that is)

sschuberth: Oh, I didn't realize @pombredanne had committed this directly, without going through a PR that runs CI. Otherwise that change should have never gotten in.

Issue #39, opened 2018-03-06
Parsing the .tag file gives error


techytushar:
When parsing the `sample.tag` file using `python parse_tv.py sample.tag`, it throws an error `TypeError: expected string or buffer`. The `sample.tag` is generated after running `python write_tv.py sample.tag`. IMHO this happens because when `write_tv.py` is executed it creates a key value pair `PackageCopyrightText: NOASSERTION`, here type of `NOASSERTION` is `<class 'spdx.utils.NoAssert'>` which is not supported. 

Whole error log can be found in the screenshot below.
![screenshot from 2018-03-06 18-07-05](https://user-images.githubusercontent.com/26834658/37032646-524ae6a8-2169-11e8-99dd-21a0f21bc571.png)

pombredanne:
Thanks for the report and PR

meretp:
@nicoweidner This issue can be closed.

PR #43, opened 2018-03-06
Fix TypeError while parsing .tag file #39

techytushar:
@pombredanne I have fixed the `TypeError` but while parsing `sample.tag` file it still gives an error `Only one PackageHomePage allowed, extra at line: 21` although there is only one key with `PackageHomePage` in the whole document.
PR for issue #39 
Signed-off-by: Tushar Mittal <chiragmittal.mittal@gmail.com>


pombredanne:
Thanks. Can you add tests for this? and also highlight the problem you mentioned as a failing tests?

pombredanne:
@techytushar ping!

PR #44, opened 2018-03-07
Add example file to convert tv to rdf

techytushar:
There was a `rdf_to_tv.py` file, i thought it would be nice to have `tv_to_rdf.py` also.
Signed-off-by: Tushar Mittal <chiragmittal.mittal@gmail.com>


examples/tv_to_rdf.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env python
+
+# Converts an tag/value file to RDF format.
```
pombredanne: Use """ docstring instead of # comments

examples/tv_to_rdf.py
```diff
@@ -0,0 +1,29 @@
...
+# Usage: tv_to_rdf <tagvaluefile> <rdffile>
+
+if __name__ == '__main__':
```
pombredanne: Could you make the code below a function and move the import at the top?
And call that function in this `if __name__ == '__main__'` block instead?

techytushar:
done the required changes, please review @pombredanne

pombredanne:
Sorry for this late merge!

PR #51, opened 2018-03-15
Return messages instead of modifying list #50

yash-nisar:
We now return messages from each `validation` function
instead of modifying lists in place.

Signed-off-by: Yash Nisar <yash.nisar@somaiya.edu>

Closes https://github.com/spdx/tools-python/issues/50

Review needed @sschuberth @pombredanne @rtgdk

### DIFF
--- a/spdx/annotation.py
+++ b/spdx/annotation.py
@@ -72,54 +72,39 @@ class Annotation(object):
     def has_comment(self):
         return self.comment is not None
 
-    def validate(self, messages=None):
+    def validate(self, messages):
         """Returns True if all the fields are valid.
         Appends any error messages to messages parameter.
         """
-        # FIXME: we should return messages instead
-        messages = messages if messages is not None else []
+        messages = self.validate_annotator(messages)
+        messages = self.validate_annotation_date(messages)
+        messages = self.validate_annotation_type(messages)
+        messages = self.validate_spdx_id(messages)
 
-        return (self.validate_annotator(messages)
-            and self.validate_annotation_date(messages)
-            and self.validate_annotation_type(messages)
-            and self.validate_spdx_id(messages))
+        return messages
 
-    def validate_annotator(self, messages=None):
-        # FIXME: we should return messages instead
-        messages = messages if messages is not None else []
+    def validate_annotator(self, messages):
+        if self.annotator is None:
+            messages = messages + ['Annotation missing annotator.']
 
-        if self.annotator is not None:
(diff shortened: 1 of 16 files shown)


sschuberth:
> We now return messages from each `validation` function instead of modifying lists in place.

Ok, but *why* is that a good thing? I prefer if commit messages do not simply repeat in prose what can be seen from the diff, but explain the *rationale* behind the change.

yash-nisar:
Sure, will add explain the rationale behind the change in a moment :) 

Thanks for your feedback.

yash-nisar:
@sschuberth Is the commit message good enough ?

sschuberth:
Thanks @yash-nisar, the message reads good now, but I'll leave the code comments to real Python-people like @pombredanne or @haikoschol.

yash-nisar:
Thanks @sschuberth :)

yash-nisar:
@pombredanne Can I have your review ? :)

pombredanne:
@yash-nisar I would be interested to see if this breaks scancode or not before merging. There were some weird cases where the messages data passed was empty. I would assume that this would fix this. 
Yet, I would be more comfortable if you could add a few new tests that would try to tests these corner cases: what happens if `messages` is None with your updates?

yash-nisar:
@pombredanne 
re
> what happens if messages is None with your updates?

Do you mean what would happen if our document was perfectly valid and the final `validate(...)` function returns an empty list ? Well, in that case, I've modified the tests by adding this line: 
`assert doc.validate([]) == []`. And in this case (where our document is perfectly valid), an empty list is passed to the fellow `validate(...)` functions and an empty list is returned by those functions.

yash-nisar:
@pombredanne Any inputs ?

yash-nisar:
@pombredanne @krysnuvadga Any inputs ?

yash-nisar:
Note: This PR will be updated after the other open PRs get merged so please **do not merge** till then.

pombredanne:
@yash-nisar you wrote
> Note: This PR will be updated after the other open PRs get merged so please do not merge till then. .... therefore I did not touched this

yash-nisar:
@pombredanne Roger that.

pombredanne:
@yash-nisar you said in a chat:

> We do return the message list which shall be empty if everything is according to our standards and shall be filled in if there are some errors

So I guess this is OK as it is now. Let me merge this then.

PR #71, opened 2018-05-25
Upgrade the 'Document' class to the SPDX 2.1 specification

yash-nisar:


spdx/document.py
```diff
@@ -189,7 +190,13 @@ class Document(object):
     Represent an SPDX document with these fields:
     - version: Spec version. Mandatory, one - Type: Version.
     - data_license: SPDX-Metadata license. Mandatory, one. Type: License.
+    - name: Name of the document. Mandatory, one. Type: str.
```
pombredanne: Do we really need to make this mandatory?

spdx/document.py
```diff
@@ -189,7 +190,13 @@ class Document(object):
...
+    - spdx_id: SPDX Identifier for the document to refer to itself in
```
pombredanne: Do we really need to make this mandatory too?

spdx/document.py
```diff
@@ -189,7 +190,13 @@ class Document(object):
...
+      relationship to other elements. Mandatory, one. Type: str.
+    - ext_document_references: External SPDX documents referenced within the
+        given SPDX document. Optional, one or many. Type: ExternalDocumentRef
     - comment: Comments on the SPDX file, optional one. Type: str
+    - namespace: SPDX document specific namespace. Mandatory, one. Type: str
```
pombredanne: Do we really need to make this mandatory again?

Note that your doc does not seem to be what happens: this does not seem mandatory in the code.

spdx/document.py
```diff
@@ -237,11 +252,14 @@ def validate(self, messages=None):
 
         return (self.validate_version(messages)
             and self.validate_data_lics(messages)
+            and self.validate_name(messages)
```
pombredanne: we likely want the validation to be strict or not.... Note that this PR can still go through and this can be fixed afterwards

spdx/document.py
yash-nisar: I had referred to https://spdx.org/sites/cpstandard/files/pages/files/spdxversion2.1.pdf#page=11 where the `Cardinality` is `Mandatory`. So, what should we do ?

spdx/document.py
yash-nisar: I had referred to https://spdx.org/sites/cpstandard/files/pages/files/spdxversion2.1.pdf#page=10 where the `Cardinality` is `Mandatory`. So, what should we do ?

spdx/document.py
yash-nisar:
```
    def validate_namespace(self, messages=None):
        # FIXME: messages should be returned
        messages = messages if messages is not None else []

        if self.namespace is None:
            messages.append('Document has no namespace.')
            return False
        else:
            return True
```
has been defined which will throw an error if the namespace is not defined :)
Do you want me to add additional tests ?

spdx/document.py
yash-nisar: I'm planning to handle the validation part after Phase 1 ? Is that okay ?

spdx/external_document_ref.py
```diff
@@ -0,0 +1,89 @@
+
```
yash-nisar: @pombredanne Do you want me to add this in `document.py` itself ?

pombredanne: Sure.

pombredanne:
This needs to be updated to resolve merge conflicts.

pombredanne:
This needs to be updated to resolve merge conflicts.

yash-nisar:
@pombredanne Merge conflicts resolved. Wanna review this one last time ?

spdx/document.py
pombredanne: leave as is.

spdx/document.py
pombredanne: that's ok for now

spdx/document.py
pombredanne: yes

spdx/document.py
```diff
@@ -268,6 +357,55 @@ def validate_data_lics(self, messages=None):
...
+
+        if self.spdx_id.endswith('SPDXRef-DOCUMENT'):
+            return True
+        else:
+            messages.append('Invalid Document SPDX Identifier value.')
+            return False
+
+    def validate_ext_document_references(self, messages=None):
+        # FIXME: messages should be returned
+        messages = messages if messages is not None else []
```
pombredanne: could a simpler `messages = messages or []` be enough here and elsewhere?

PR #103, opened 2019-04-17
Fix #102 by using Version method from_str in config.py file.

### LINKED ISSUE #102: Version model fields must be integers but sometimes they aren't
**BACKGROUND INFORMATION**
`major` and `minor` fields of `Version` model are supposed to be integers and the class method `Version.from_str()` is responsible for assuring that.
https://github.com/spdx/tools-python/blob/301d72f6ae57c832c1da7f6402fa49b192de6810/spdx/version.py#L36-L45
If `license_list_version` field is absent when initializing a `CreationInfo` object, a default value is taken from the `config.py` file.
https://github.com/spdx/tools-python/blob/301d72f6ae57c832c1da7f6402fa49b192de6810/spdx/creationinfo.py#L131-L136
**THE ISSUE**
The default `Version` object described above is created assigning `major` and `minor` fields directly as string values.
https://github.com/spdx/tools-python/blob/301d72f6ae57c832c1da7f6402fa49b192de6810/spdx/config.py#L48-L49
`__lt__` method and parsers tests (expect integer values) may fail.

xavierfigueroav:
Fix #102 
This PR modifies config.py file to use the method `from_str` in `Version` instead of creating the object directly with string values.

Signed-off-by: Xavier Figueroa <xavierfigueroav@gmail.com>

### DIFF
--- a/spdx/config.py
+++ b/spdx/config.py
@@ -34,7 +34,7 @@ def load_license_list(file_name):
     licenses_map = {}
     with codecs.open(file_name, 'rb', encoding='utf-8') as lics:
         licenses = json.load(lics)
-        version = licenses['licenseListVersion'].split('.')
+        version = licenses['licenseListVersion']
         for lic in licenses['licenses']:
             if lic.get('isDeprecatedLicenseId'):
                 continue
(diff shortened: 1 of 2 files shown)


xavierfigueroav:
> We would need some tests for this!
@pombredanne There are some tests now!

pombredanne:
But this needs to be rebased too

xavierfigueroav:
> All good. Thank you!

Great!

> But this needs to be rebased too

@pombredanne It is now rebased!

pombredanne:
Thanks!

PR #76, opened 2018-07-06
Add 'Snippet' class as a part of the SPDX 2.1 specification

yash-nisar:
Add 'spdx_id', 'name', 'comment', 'copyright', and
'license_comment' to the 'snippet' class.

Signed-off-by: Yash Nisar <yash.nisar@somaiya.edu>

PR #107, opened 2019-06-07
Add 'Snippet' class as a part of the SPDX 2.1 specification

pombredanne:
This is to add 'Snippet' class as a part of the SPDX 2.1 specification
This is https://github.com/spdx/tools-python/pull/76 from @yash-nisar rebased by @xavierfigueroav


pombredanne:
LGTM .... Thanks!

PR #110, opened 2019-06-12
 Fix #91 by passing all rdflib types in six.text_type function

### LINKED ISSUE #91: Data from parsers should be stored in spdx models only as python types and spdx models
When parsing documents from _RDF_ files, some fields are being stored as _rdflib_ objects and it can be problematic.

All data stored in SPDX models (_document.Document_, _file.File_, _document.ExtractedLicense_, etc) should be stored as _python_ and _spdx_ types so that they are independent of the format to be written or read.

I had problems trying to write _YAML_ files from a document instance created by reading an _RDF_ file because _pyyaml_ was not able to represent some _rdflib_ objects (and it does not have to be able to). As expected, _**py**yaml_ is able to deal with python objects.

Of course, almost every _spdx_ type has overwritten the _\_\_str\_\__ and _\_\_repr\_\__ methods and some others have methods that return the 'str' representation of their single fields, but they sometimes assume (correctly) that fields such as full_name, text or comment (from _document.ExtractedLicense_) are strings (but they are _rdflib.terms.Literal_ objects when parsed from _RDF_ files), so those functions just return the field (_rdflib_ type).

Again, _spdx_ types should be independent of the format to be written or read.

xavierfigueroav:
1. About fixing:
 - Pass all rdflib types in RDF parsers into six.text_type function before passing them into RDF builders.
2. About tests:
 - Add test utilities that build a Python dictionary representation of the information stored in models.
 - Add a JSON file with expected info after parsing a test file.
 - Add tests that open a test file, open the expected file and compare them.

Signed-off-by: Xavier Figueroa <xavierfigueroav@gmail.com>

### DIFF
--- a/data/SPDXRdfExample.rdf
+++ b/data/SPDXRdfExample.rdf
@@ -118,7 +118,6 @@
     <referencesFile>
       <File rdf:about="https://spdx.org/spdxdocs/spdx-example-444504E0-4F89-41D3-9A0C-0305E82C3301#SPDXRef-File2">
         <copyrightText>Copyright 2010, 2011 Source Auditor Inc.</copyrightText>
-        <licenseComments></licenseComments>
         <licenseInfoInFile rdf:resource="http://spdx.org/licenses/Apache-2.0"/>
         <licenseConcluded rdf:resource="http://spdx.org/licenses/Apache-2.0"/>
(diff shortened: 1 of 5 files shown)


xavierfigueroav:
A test is currently failing due to the change in the license list version format. Waiting until #103 is merged.

pombredanne:
Thanks!

PR #114, opened 2019-08-21
Fix parsing of PackageHomePage (was probably a copy and past error),

akleber:
add PackageHomePage tag to SPDXTagExample.tag

Signed-off-by: Andreas Kleber <andreas@drosselweg7a.de>


pombredanne:
@akleber Thank you ++

Issue #113, opened 2019-08-21
Add AUTHORS file


xavierfigueroav:
An authors file is needed to list and thank people who have contributed to the project.

SKalt:
Hi, could you quote and link docs on the AUTHORS file standard?  I'm interested in what advantages the split-out AUTHORS file has over the current `authors=[...]` field in `setup.py` or extracting authors from the `git log`.

SKalt:
More googling yielded https://opensource.google/docs/releasing/authors/, which outlines how an AUTHORS file (1) delineates who has copyrights to the open-source project and (2) creates a central place to acknowledge contributors' work.  Seems worth a look to me. Then again, `git log --format=format:'%aN <%aE>' | awk '!x[$0]++'` or whatever powers Github's repo insights also can shows who committed, and possibly when and how much.

xavierfigueroav:
AUTHORS file may be a platform-independent (git/github) alternative to acknowledge contributors' work. It is also a common, good practice in open source projects.

It looks like no SPDX repository has an AUTHORS file, so, if you are interested in, you could follow the style of the [license-expression library](https://github.com/nexB/license-expression)'s one, (which is maintained by `tools-python`'s maintainer, who, by the way, suggested adding this file).

b-thebest:
Anything remaining in this issue, which I can do?

thealphacod3r:
@xavierfigueroav is this resolved or should i complete this?

harshkasat:
@meretp Is this issue still open? It has the label of 'tech debt'.

PR #124, opened 2019-10-21
added an authors file

SKalt:
Adressesses #113.  Authors are added in order of the number of commits contributed from most to least, with recency breaking ties. I used each contributor's most recent email address.

<details><summary>Code used to generate authors list</summary>

```python
from subprocess import run, PIPE
from collections import Counter, OrderedDict


def get_authors():
    return run(
        "git log --use-mailmap --format=format:'%aN#<%aE>'".split(" "),
        stdout=PIPE,
    ).stdout.decode()


def get_authors_list():
    return [auth.split("#") for auth in get_authors().split("\n")]


def by_recency(authors):
    order = OrderedDict()
    for name, email in authors:
        result = order.get(name, [])
        result.append(email)
        order[name] = result
    return "\n".join([f"{i} {order[i][0]}" for i in order])


def by_commit_count(authors):
    count = Counter()
    name_emails = {}
    for name, email in authors:
        count[name] += 1
        if name not in name_emails:
            name_emails[name] = email
    by_count = [*count.items()]
    by_count.sort(key=lambda x: x[1])
    by_count = by_count[::-1]
    return "\n".join(
        " ".join((name, name_emails[name])) for name, count in by_count
    )


if __name__ == "__main__":
    with open("AUTHORS", "w") as target:
        target.write(by_commit_count(get_authors_list()))

```
</details>

Signed-off-by: Steven Kalt <kalt.steven@gmail.com>

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,0 +1,29 @@
+# This is the list of [Project Name] authors for copyright purposes. This does
+# not necessarily list everyone who has contributed code, since in some cases,
+# their employer may be the copyright holder. To see the full list of
+# contributors, see the revision history in source control.
+
+Ahmed H. Ismail <ahm3d.hisham@gmail.com>
+Philippe Ombredanne <pombredanne@nexb.com>
+Sebastian Schuberth <sschuberth@users.noreply.github.com>
+Xavier Figueroa <xavierfigueroav@gmail.com>
+yash-nisar <yash.nisar@somaiya.edu>
+Carmen Bianca Bakker <carmen@carmenbianca.eu>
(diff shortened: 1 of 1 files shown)


AUTHORS
```diff
@@ -0,0 +1,29 @@
+# This is the list of [Project Name] authors for copyright purposes. This does
```
jayvdb, line 1: `[Project Name]` should be "spdx/tools-python" or similar.

Also it would be great if you could add the script used to create this into this PR, so it can be re-run as part of the release process.

jayvdb (changes requested):
needs to be re-run

Issue #88, opened 2019-03-20
has_optional_field implementation may lead to unexpected results


xavierfigueroav:
_has_optional_field(field)_ verifies whether the value of _field_ is not None. But, some attributes are sometimes initialized as some default/empty object, like an empty list. In that example, _has_optional_field(field)_ will always return True, even if the list is empty (when I think it should return False because there is actually no information).

I suggest: return bool(field)

b-thebest:
Can I work on this issue ?

PR #132, opened 2020-03-20
Possible solution of issue #88

kbermude:
Changing _return bool(field)_ in has_optional_field function, we will obtain is a field is optional, does not matters is the field is initialized empty, because is optional.

### DIFF
--- a/spdx/package.py
+++ b/spdx/package.py
@@ -299,7 +299,7 @@ class Package(object):
         return sha1.hexdigest()
 
     def has_optional_field(self, field):
-        return getattr(self, field, None) is not None
+        return bool (getattr(self, field, None)
 
 
 class ExternalPackageRef(object):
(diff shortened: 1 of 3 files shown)


spdx/package.py
```diff
@@ -261,7 +261,7 @@ def calc_verif_code(self):
         return sha1.hexdigest()
 
     def has_optional_field(self, field):
-        return getattr(self, field, None) is not None
+        return bool (getattr(self, field, None)
```
ShubhamKJha, line 264:
```suggestion
        return bool (getattr(self, field, None))
```

armintaenzertng (approved):

armintaenzertng:
Had to create a new PR due to sign-off issues: https://github.com/spdx/tools-python/pull/252

PR #133, opened 2020-03-30
Fix write_tv example

SantiagoTorres:
I've been fiddling with the python library lately. I found a couple of little
hiccups in the write_tv example. I'm opening this PR so I don't forget,
although I'll admit I haven't done my due dilligence figuring out if this
example is in fact already fixed somewhere else (and apologies for that!).

Also, the spdx_id fields that I chose may prove to be not entirely compliant,
as I just read the validation code and  picked something that passed...

### DIFF
--- a/examples/write_tv.py
+++ b/examples/write_tv.py
@@ -21,7 +21,10 @@ if __name__ == '__main__':
 
     doc = Document()
     doc.version = Version(1, 2)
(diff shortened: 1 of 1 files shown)


MaciejKucia:
It would be nice to have this merged as the example does not work.

pombredanne:
Merged at last! thanks.

Issue #136, opened 2020-06-08
Changing the old-style f-string (str.format() ) to new style f-strings.


Yash-Varshney:
The f-strings used in the code are very old and quite verbose when you are dealing with multiple parameters and longer strings.
Python3.6 introduced a new style f-string which are not only easy to use but also easily readable and faster than the old python2.6 one.

Yash-Varshney:
This issue is currently paused and will be a part of a bigger project in which we need to halt v2.x of python and completely update the whole repo to work with v3 and above of python.

RishabhBhatnagar:
Can you provide some benchmarks to prove the efficiency of fstrings over str.format?
Comparing fstrings with str.format of python2.7 won't be a fair comparison as many of the libraries in python3.x were rewritten in C to improve their runtime.

How does fstrings compare to str.format of python 3.x? If the difference is not significant, we can be better off without any modifications.

One more problem I can see with the replacement is that In order to provide backward compatibility to python versions < 3.6, we will be required to import functions from \_\_future__ which will be costlier than the overhead(if any) of using str.format.

zvr:
Python 3.6 is the earliest still supported version of Python.

PR #137, opened 2020-06-08
[WIP] changing the old fstring to new one.

Yash-Varshney:
Signed-off-by: Yash Varshney <b18038@students.iitmandi.ac.in>
Issue : #136 
Made changes ro one file till now. Will commit more if permissible.

### DIFF
--- a/examples/parse_json.py
+++ b/examples/parse_json.py
@@ -14,31 +14,31 @@ if __name__ == '__main__':
     with open(file) as f:
         doc, error = p.parse(f)
         if not error:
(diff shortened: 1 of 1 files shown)


RishabhBhatnagar:
The new type of fstring is introduced in python3.6. It won't work for versions less than v3.6.
python3.5 is a relatively stable release which is still used by many codebases for which this code will break. It would not be a preferable alternative to move to a newer alternative which is not backward compatible.

Yash-Varshney:
thanks, @RishabhBhatnagar for addressing the problem. I'll close this P.R. as for now.

PR #141, opened 2020-06-16
Fixes the install error in Python 2.7 and CircleCI error with Python 3.4.8

### LINKED ISSUE #140: Build failures occurring in Python 2.7
The tests are failing for Python 2.7 (build failures). This is hindering the testing of PRs.
Apparently, the current master was last tested in August 2019.

ShubhamKJha:
Signed-off-by: Shubham Kumar Jha <skjha832@gmail.com>

Fixes #140 
Build failures were occurring in Python2 due to `rdflib` allowing any version of `pyparsing` (it's dependency) to be installed. It eventually installs the latest version which doesn't work with Python 2.7.

Fix:
* Updated setup.py file to install pyparsing <= 1.5.7 when installed with Python 2.7
* Since, rdflib allows any version of pyparsing to be installed, the latest version is not installed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,7 +31,7 @@ jobs:
         brew update
         python --version
         sudo -H pip install --upgrade virtualenv
-        brew install pyenv
+        brew install pyenv -v 1.2.13
         echo 'eval "$(pyenv init -)"' >> ~/.bash_profile
         source ~/.bash_profile
         pyenv install 3.4.8
--- a/setup.py
+++ b/setup.py
@@ -30,11 +30,13 @@ setup(
     test_suite='setup.test_suite',
     install_requires=[
         'ply',
+        'pyparsing<=1.5.7;python_version<="2.8"',
         'rdflib',
         'six',
         'pyyaml',
         'xmltodict',
     ],
+    python_requires='>=2.7',
     entry_points={
         'console_scripts': [
             'spdx-tv2rdf = spdx.tv_to_rdf:main',


ShubhamKJha:
@pombredanne I've fixed the issues with Python 2.7 

The build is but failing for Python 3.4.8 (apparently as the error logs say, brew is not able to install it), I've made some changes but I can't get the build started for Circle CI. 
@pombredanne can you start the builds.

ShubhamKJha:
The CircleCI error happening with Python 3.4.8 is due to the updated version of pyenv (used to install different Python versions) which has a dependency on openssl-1.1. The support for openssl-1.1 is provided by Pyhon 2.7.13 and Python 3.5 +.

Fix: Pinning the version of pyenv (1.2.13) which requires openssl-1.0 fixed the problem.

ping @goneall @rtgdk @pombredanne Please, check this PR and merge.

pombredanne:
note that we do not want to support Python 3.4, actually going forward any version under 3.6 should be dropped.

ShubhamKJha:
Dropping support for older Python version would be great. I just felt that we may need to use this in spdx-online-tools, which currently uses Python 2.7, so it must be installable with that.

pombredanne:
> Dropping support for older Python version would be great. I just felt that we may need to use this in spdx-online-tools, which currently uses Python 2.7, so it must be installable with that.

2.7 is obsolete, so spdx-online-tools would have to stick using older versions IMHO.

goneall:
There's work in progress to upgrade the SPDX online tools to Python3: https://github.com/spdx/spdx-online-tools/pull/178

Issue #123, opened 2019-10-19
Review and possibly update JSON and YAML file generation to match the SPDX 2.2 spec proposals


goneall:
Several changes to the JSON and YAML formats were discussed and generally agreed on for the SPDX 2.2 spec.

There is a PR with changes to the example file: https://github.com/spdx/spdx-spec/pull/149.  The PR documents the related issues which were resolved (sorry for the extra clicks to find get all the documentation).

The Python libraries may need to be updated to match the 2.2 spec.

Yash-Varshney:
Hi @goneall . I am interested in updating `spdx-python-tools` to support spdx v2.2
Can you describe a little bit more about the problem and things needed to be updated further?
Thanks for your time.

goneall:
One suggested approach is to convert the [tag/value example file](https://github.com/spdx/spdx-spec/blob/development/v2.2.1/examples/SPDXTagExample-v2.2.spdx) to a JSON using the Python libraries and compare that to the [JSON example file](https://github.com/spdx/spdx-spec/blob/development/v2.2.1/examples/SPDXJSONExample-v2.2.spdx.json).  Analyzing the difference should provide a guide as to what needs to be updated.

nicoweidner:
I used [the 2.2 tag example from spdx-spec](https://github.com/spdx/spdx-spec/blob/v2.2.2/examples/SPDXTagExample-v2.2.spdx) and converted it to json using the python tools. The converted file is attached.

Using the json-based comparison tooling from spdx-testbed, the following differences were detected:
```
[ListDifference(actualValue="Organization: ExampleCodeInspect", expectedValue=null, path=/creationInfo/creators/1, pathInReferenceDoc=/creationInfo/creators, comment=No element in expected list with a matching Spdx id or no Spdx id present.),
    ListDifference(actualValue="Person: Jane Doe", expectedValue=null, path=/creationInfo/creators/2, pathInReferenceDoc=/creationInfo/creators, comment=No element in expected list with a matching Spdx id or no Spdx id present.),
    ListDifference(actualValue=null, expectedValue="Organization: ExampleCodeInspect ()", path=/creationInfo/creators, pathInReferenceDoc=/creationInfo/creators/1, comment=No element in actual list with a matching Spdx id or no Spdx id present.),
[...]

PR #144, opened 2020-08-11
CommunityBridge Task1 + Task2 combined P.R.

Yash-Varshney:
This P.R. is for evaluation purpose. **DO NOT MERGE**
`Task 1` was about adding relationship class.
In `Task 2` I've added attribution text in file, snippet and package class for all the formats in builders, parsers and writers. I have also changed the name of `namespace` to `documentNamespace` [issue #123 ]and `spdxDocumentNamespace` to `spdxDocument` in externalDocs for JSON, YAML and XML formats and made all the desired changes. 
Everything is working perfectly and I have also created tests for everything needed.

Please Ignore many small changes which are caused due to the use of `BLACK` formatter for all the python code. Also, I have not rebased and squashed it yet, because it will be easy to differentiate between task 1 and task 2 code [for easily accessing you can just click the commits from the 10th]. I couldn't make a P.R. separately for task1 and task2 because there are interconnected commits [ Task 3 will be independent of these both].

### DIFF
--- a/spdx/relationship.py
+++ b/spdx/relationship.py
@@ -14,7 +14,17 @@ from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import unicode_literals
 
-from enum import Enum, auto
+from enum import Enum
+
+# Implement the auto feature that becomes available in 3.6	
+autoinc = 0	
+
+
+def auto():	
+    global autoinc	
+    autoval = autoinc	
+    autoinc += 1	
+    return autoval
 
 
 class RelationshipType(Enum):
--- a/spdx/relationship.py
+++ b/spdx/relationship.py
@@ -1,0 +1,122 @@
+# Copyright (c) 2020 Yash Varshney
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#    http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import absolute_import
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from enum import Enum
+
+# Implement the auto feature that becomes available in 3.6
+autoinc = 0
+
+
+def auto():
+    global autoinc
+    autoval = autoinc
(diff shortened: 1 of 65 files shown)


Yash-Varshney:
also, all the project details can be easily accessed from project_details.md file in the P.R.

spdx/relationship.py
```diff
@@ -0,0 +1,122 @@
...
+# Implement the auto feature that becomes available in 3.6
+autoinc = 0
+
+
+def auto():
+    global autoinc
+    autoval = autoinc
+    autoinc += 1
+    return autoval
+
```
rtgdk, line 28: We could also use enum.auto() function from enum module, instead of creating our own.

Yash-Varshney: The `auto` feature is supported only after 3.6 version of python and presently python-tools does not support it. i.e. why I made the same auto feature here.

Yash-Varshney:
> @Yash-Varshney Thank you for opening the PR. It took me a while to find the actual code changes apart from linting changes 😅 . Left a comment. Other than that PR looks good to me. You can delete this PR if you want.

Thanks @rtgdk for reviewing.

spdx/relationship.py
```diff
@@ -58,53 +90,7 @@ def validate(self, messages):
...
-            "PROVIDED_DEPENDENCY_OF",
-            "RUNTIME_DEPENDENCY_OF",
-            "STATIC_LINK",
-            "TEST_CASE_OF",
-            "TEST_DEPENDENCY_OF",
-            "TEST_TOOL_OF",
-            "VARIANT_OF",
-        ]:
-
+        if r_type not in [name for name, _ in RelationshipType.__members__.items()]:
```
RishabhBhatnagar, line 115: Instead of building a new list for storing all the enum-members, 
we could just have used the dictionary keys.

`RelationshipType.__members__` is a dictionary.

`if r_type not in RelationshipType.__members__.keys()` would be a preferred way for the following reasons:
1. `__members__` is a dictionary object, it's keys method returns a dictview object which takes amortized O(1) complexity to return a view object as opposed to O(N) for building a new list for all the members.
2. It is much more clean and shorter than the current method being used. The current method unnecessarily creates a new list object which consumes additional space and time.

Some might argue that we can even check membership using `RelationshipType._member_names_`. The reason I don't suggest using `_member_names` attribute is that I couldn't find that attribute in the latest official python documentation. Otherwise, it could've been more intuitive way to check membership.

Yash-Varshney: thanks for your advice Rishabh, could you please see this [issue](https://github.com/Yash-Varshney/tools-python/issues/2) which is the inspiration for this code.

Issue #149, opened 2020-10-11
when files_analyzed is set to false, validation does not comply with specs


alpianon:
According to section 3.8.1, 3.8.3, 3.8.4, 3.9.3 and 3.10.3 of SPDX 2.2 specs:

> ## 3.8 Files Analyzed
> 3.8.1 Purpose: Indicates whether the file content of this package has been available for or subjected to analysis when creating the SPDX document. If false, indicates packages that represent metadata or URI references to a project, product, artifact, distribution or a component. **If false, the package must not contain any files**.
>
> [...]
>
> 3.8.3 Cardinality: **Optional, one. If omitted, the default value of true is assumed.**
> 
> 3.8.4 Data Format: **Boolean**
>
> [...]
>
> ## 3.9 Package Verification Code
> [...]
>
> 3.9.3 Cardinality: Mandatory, one if FilesAnalyzed is true or omitted, zero (**must be omitted) if FilesAnalyzed is false**.
>
> [...]
>
> ## 3.10 Package Checksum
> 3.10.1 Purpose: Provide an independently reproducible mechanism that permits unique identification of a specific package that correlates to the data in this SPDX file. This identifier enables a recipient to determine if any file in the original package has been changed.  [...]
> [...]
>
> 3.10.3 Cardinality: **Optional**, one or many.

So, **when files_analyzed is false** in a package:

- there must be no files in the package, so **files must not be validated** (section 3.8.1)
- there must be no **verif_code**, so it **must not be validated** (section 3.9.3)
- there may be no **check_sum** (which is optional, anyway - section 3.10.3), so it **should be validated only when it is not None/omitted**

And **files_analyzed should be validated, too (must be True, False or None** (None="omitted"))

The current **validation methods** of the Package class **do not comply with any of the above**, because:

- they try to validate files even when files_analyzed is false and no file must be included in the SPDX doc
- they try to validate verif_code as mandatory field even when files_analyzed is false and verif_code must not be included
- they always try to validate check_sum, even if it is an optional field

This leads to **validation errors even if the document fully complies with the specs**, and, on the other hand, to validation of non-compliant documents.

Finally, in `tagvaluebuilder.py`, **files_analyzed** is stored as string in the package object, while it **should be stored as boolean**.

alpianon:
Also package checksum validation should be skipped when there are no files analyzed (and in any case it is an optional field, according to the specs, so it should not be validated when it is set to None)

alpianon:
[...]

PR #150, opened 2020-10-11
fix validation errors when files_analyzed is false

alpianon:
( solves issue #149 )

According to **SPDX 2.2 specs**, when files_analyzed is false in a package:

- there *must* be *no* files in the package, so files must *not* be validated (**section 3.8.1**)
- there *must* be *no* verif_code (**section 3.9.3**)
- there may be no check_sum (which is optional, anyway - **section 3.10.3**)

Finally, also files_analyzed should be validated (must be True, False or None, (None="omitted")

This patch covers all the above, and fixes an error in tagvaluebuilder.py, where files_analyzed was stored as string in the package object, while it should be stored as boolean.

Moreover, rdf and tagvalue writers, as well as test files, have been patched accordingly.

### DIFF
--- a/spdx/package.py
+++ b/spdx/package.py
@@ -43,7 +43,8 @@ class Package(object):
      If set to "false", the package must not contain any files.
      Optional, boolean.
      - homepage: Optional, URL as string or NONE or NO_ASSERTION.
-     - verif_code: Mandatory string.
+     - verif_code: string. Mandatory if files_analyzed is True or None (omitted)
+     Must be None (omitted) if files_analyzed is False
      - check_sum: Optional , spdx.checksum.Algorithm.
      - source_info: Optional string.
      - conc_lics: Mandatory spdx.document.License or spdx.utils.SPDXNone or
@@ -107,6 +108,7 @@ class Package(object):
         Validate the package fields.
         Append user friendly error messages to the `messages` list.
         """
+        messages = self.validate_files_analyzed(messages)
         messages = self.validate_checksum(messages)
         messages = self.validate_optional_str_fields(messages)
         messages = self.validate_mandatory_str_fields(messages)
@@ -114,6 +116,18 @@ class Package(object):
         messages = self.validate_pkg_ext_refs(messages)
         messages = self.validate_mandatory_fields(messages)
         messages = self.validate_optional_fields(messages)
+
+        return messages
+
+    def validate_files_analyzed(self, messages):
+        if self.files_analyzed not in [ True, False, None ]:
+            messages = messages + [
+                'Package files_analyzed must be True or False or None (omitted)'
+            ]
+        if self.files_analyzed == False and self.verif_code is not None:
+            messages = messages + [
+                'Package verif_code must be None (omitted) when files_analyzed is False'
+            ]
 
         return messages
 
@@ -176,13 +190,14 @@ class Package(object):
         return messages
 
     def validate_files(self, messages):
-        if not self.files:
-            messages = messages + [
-                'Package must have at least one file.'
-            ]
-        else:
-            for f in self.files:
-                messages = f.validate(messages)
+        if self.files_analyzed != False:
+            if not self.files:
+                messages = messages + [
+                    'Package must have at least one file.'
+                ]
+            else:
+                for f in self.files:
+                    messages = f.validate(messages)
 
         return messages
 
@@ -207,7 +222,9 @@ class Package(object):
         """Fields marked as Mandatory and of type string in class
         docstring must be of a type that provides __str__ method.
         """
-        FIELDS = ['name', 'spdx_id', 'download_location', 'verif_code', 'cr_text']
+        FIELDS = ['name', 'spdx_id', 'download_location', 'cr_text']
+        if self.files_analyzed != False:
+            FIELDS = FIELDS + ['verif_code']
         messages = self.validate_str_fields(FIELDS, False, messages)
 
         return messages
@@ -233,13 +250,14 @@ class Package(object):
         return messages
 
     def validate_checksum(self, messages):
-        if not isinstance(self.check_sum, checksum.Algorithm):
-            messages = messages + [
-                'Package checksum must be instance of spdx.checksum.Algorithm'
-            ]
-        else:
-            if self.check_sum.identifier != 'SHA1':
-                messages = messages + ['File checksum algorithm must be SHA1']
+        if self.check_sum is not None: # optional
+            if not isinstance(self.check_sum, checksum.Algorithm):
+                messages = messages + [
+                    'Package checksum must be instance of spdx.checksum.Algorithm'
+                ]
+            else:
+                if self.check_sum.identifier != 'SHA1':
+                    messages = messages + ['File checksum algorithm must be SHA1']
 
         return messages
 
--- a/spdx/package.py
+++ b/spdx/package.py
@@ -250,7 +250,7 @@ class Package(object):
         return messages
 
     def validate_checksum(self, messages):
-        if self.check_sum is not None: # optional
+        if self.check_sum is not None:
             if not isinstance(self.check_sum, checksum.Algorithm):
                 messages = messages + [
                     'Package checksum must be instance of spdx.checksum.Algorithm'
--- a/spdx/package.py
+++ b/spdx/package.py
@@ -124,7 +124,7 @@ class Package(object):
             messages = messages + [
                 'Package files_analyzed must be True or False or None (omitted)'
             ]
-        if self.files_analyzed == False and self.verif_code is not None:
+        if self.files_analyzed is False and self.verif_code is not None:
             messages = messages + [
                 'Package verif_code must be None (omitted) when files_analyzed is False'
             ]
--- a/spdx/parsers/tagvaluebuilders.py
+++ b/spdx/parsers/tagvaluebuilders.py
@@ -683,8 +683,10 @@ class PackageBuilder(object):
             if files_analyzed:
                 if validations.validate_pkg_files_analyzed(files_analyzed):
                     self.package_files_analyzed_set = True
-                    doc.package.files_analyzed = files_analyzed
-                    print(doc.package.files_analyzed)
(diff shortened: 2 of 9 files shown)


goneall (approved):
Changes LGTM

goneall:
@alpianon Thanks for these fixes 👍 

We just merged in some work done which unfortunately caused some merge conflicts.

Could you resolve the conflicts?

@zvr @rtgdk @pombredanne Could one of you do a second review to meet the merge criteria?

alpianon:
Basically the conflict arose because single quotes have been converted to double quotes in master branch... now it should work. However, some CI pipelines are broken and some checks fail...

spdx/package.py
```diff
@@ -127,6 +129,18 @@ def validate(self, messages=None):
 
         return messages
 
+    def validate_files_analyzed(self, messages):
+        if self.files_analyzed not in [ True, False, None ]:
+            messages = messages + [
+                'Package files_analyzed must be True or False or None (omitted)'
+            ]
+        if self.files_analyzed == False and self.verif_code is not None:
```
pombredanne, line 137:
```suggestion
        if self.files_analyzed is False and self.verif_code:
```

spdx/package.py
```diff
@@ -194,11 +208,12 @@ def validate_mandatory_fields(self, messages):
         return messages
 
     def validate_files(self, messages):
-        if not self.files:
-            messages = messages + ["Package must have at least one file."]
-        else:
-            for f in self.files:
-                messages = f.validate(messages)
+        if self.files_analyzed != False:
```
pombredanne, line 211:
```suggestion
        if self.files_analyzed is not False:
```

spdx/package.py
```diff
@@ -194,11 +208,12 @@ def validate_mandatory_fields(self, messages):
...
+            if not self.files:
+                messages = messages + ["Package must have at least one file."]
```
pombredanne, line 213: Nit: what about using single quotes there for consistency or using double quotes everywhere?

spdx/writers/rdf.py
```diff
@@ -796,7 +796,8 @@ def create_package_node(self, package):
         )
         self.graph.add(down_loc_node)
         # Handle package verification
-        verif_node = self.package_verif_node(package)
+        if package.files_analyzed != False:
```
pombredanne, line 799:
```suggestion
        if package.files_analyzed is not False:
```

tests/test_document.py
```diff
@@ -102,7 +103,7 @@ def test_document_validate_failures_returns_informative_messages(self):
             'Package declared license must be instance of spdx.utils.SPDXNone '
             'or spdx.utils.NoAssert or spdx.document.License'
         ]
-        assert expected == messages
+        assert sorted(expected) == sorted(messages)
```
pombredanne, line 106: I wonder why the results order would not be stable?

spdx/parsers/tagvaluebuilders.py
```diff
@@ -724,8 +724,10 @@ def set_pkg_files_analyzed(self, doc, files_analyzed):
             if files_analyzed:
                 if validations.validate_pkg_files_analyzed(files_analyzed):
                     self.package_files_analyzed_set = True
-                    doc.package.files_analyzed = files_analyzed
-                    print(doc.package.files_analyzed)
+                    doc.package.files_analyzed = (files_analyzed.lower() == 'true')
+                    # convert to boolean;
```
pombredanne, line 728: May be you meant to remove these comments?

spdx/package.py
```diff
@@ -248,13 +265,14 @@ def validate_str_fields(self, fields, optional, messages):
...
 
     def validate_checksum(self, messages):
-        if not isinstance(self.check_sum, checksum.Algorithm):
-            messages = messages + [
-                "Package checksum must be instance of spdx.checksum.Algorithm"
-            ]
-        else:
-            if self.check_sum.identifier != "SHA1":
-                messages = messages + ["File checksum algorithm must be SHA1"]
+        if self.check_sum is not None: # optional
```
pombredanne, line 268: Can you avoid end of line comments?... also I am not sure what this optional means here?
And why not using instead:
```suggestion
        if self.check_sum:
```

spdx/package.py
```diff
@@ -224,7 +239,9 @@ def validate_mandatory_str_fields(self, messages):
         """Fields marked as Mandatory and of type string in class
         docstring must be of a type that provides __str__ method.
         """
-        FIELDS = ["name", "spdx_id", "download_location", "verif_code", "cr_text"]
+        FIELDS = ["name", "spdx_id", "download_location", "cr_text"]
+        if self.files_analyzed != False:
+            FIELDS = FIELDS + ["verif_code"]
```
pombredanne, line 244:
```suggestion
            FIELDS.append("verif_code")
```

spdx/package.py
```diff
@@ -224,7 +239,9 @@ def validate_mandatory_str_fields(self, messages):
         """Fields marked as Mandatory and of type string in class
         docstring must be of a type that provides __str__ method.
         """
-        FIELDS = ["name", "spdx_id", "download_location", "verif_code", "cr_text"]
+        FIELDS = ["name", "spdx_id", "download_location", "cr_text"]
+        if self.files_analyzed != False:
```
pombredanne, line 243:
```suggestion
        if not self.files_analyzed:
```

pombredanne (commented):
Thanks!
I have added a few nitpickings and comments for you review.
Also (globally as this a style that already was there), I find appending to a list more readable rather than doing an addition of a new list. e.g
- `messages.append('some message')` is IMHO more concise, explicit and readable than:
- `messages += ['some message']` which is slight improvement over the original:
- `messages = messages + ['some message']`

pombredanne:
> Basically the conflict arose because single quotes were converted to double quotes in master branch... now it should work. However, some CI pipelines are broken and some checks fail...

These are broken and need to be reformed entirely indeed

spdx/package.py
alpianon:
```suggestion
        if self.files_analyzed is False and self.verif_code is not None:
```

I would keep the check that verif_code "is not None": if someone, by mistake, puts an empty string instead of omitting the tag, I think that it would better *not* to validate the file. What do you think?

spdx/package.py
alpianon: ok

tests/test_document.py
alpianon: In all my tests the result order was always stable, but I guess it does not hurt to add `sorted` here...

spdx/writers/rdf.py
alpianon: ok

spdx/package.py
alpianon: sorry for the inline comment :)  optional means that check_sum tag, according to the specs, is optional (not mandatory). But it is already written in Package class' help text, so it is not necessary to repeat it here.

As for keeping "is not None", the same reasoning applies as before for verif_code

```suggestion
        if self.check_sum is not None:
```

spdx/package.py
alpianon: Is this the right line for this comment? I just realized that I forgot to change quotes at lines 135 and 139, it this what you mean?

spdx/package.py
alpianon: This is on purpose. Specs say: 

> ## 3.8 Files Analyzed 
>
> [...]
>
> 3.8.3 Cardinality: Optional, one. **If omitted, the default value of true is assumed**.

So if files_analyzed is None, we have to assume that its value is True and do the check

spdx/package.py
alpianon: ok

spdx/parsers/tagvaluebuilders.py
alpianon: It was just a reminder :)

Please remove them

alpianon (commented):
I closed most of the conversations, some remain open (see my inline comments)

spdx/package.py
alpianon:
```suggestion
        if self.files_analyzed != False:
```

spdx/package.py
pombredanne: The pythonic way would be IMHO:
`if not self.files_analyzed is False and self.verif_code:`

pombredanne (changes requested):
There are just a few quotes issues and a question for you.
Also we enabled the DCO signoff recently... do you mind adding it at least in a message with your signoff in this PR?

spdx/parsers/tagvaluebuilders.py
alpianon:
```suggestion
                    doc.package.files_analyzed = (files_analyzed.lower() == 'true')
```

alpianon:
```suggestion
```

spdx/package.py
alpianon:
```suggestion
        if self.files_analyzed != False:
            # FIXME: for clarity and better readability, self.files_analyzed should 
            # default to True when the corresponding value is omitted in the SPDX file
            # (see SPDX specs 3.8.3)
```

alpianon: is it OK if we put a reminder here, and then I will make another PR?

alpianon:
> There are just a few quotes issues and a question for you.
> Also we enabled the DCO signoff recently... do you mind adding it at least in a message with your signoff in this PR?

This should be solved now, please check

alpianon:
> > There are just a few quotes issues and a question for you.
> > Also we enabled the DCO signoff recently... do you mind adding it at least in a message with your signoff in this PR?
> 
> This should be solved now, please check

Strange: in this conversation page the DCO check is shown as failing, but if you click on "Details", you see that the DCO check is OK...

pombredanne:
> Strange: in this conversation page the DCO check is shown as failing, but if you click on "Details", you see that the DCO check is OK...

I fored it as passing :)

pombredanne:
This has been merged after a rebase.
Thanks!

PR #156, opened 2021-02-24
Moved the metadata into setup.cfg.

KOLANICH:
Added pyproject.toml.
Renamed console_scripts because their names are too generic to have this package installed into system.
Versions are now fetched from git tags.


setup.cfg
```diff
@@ -1,8 +1,49 @@
...
+license = Apache-2.0
 license_file = LICENSE
+description = SPDX parser and tools.
+url = https://github.com/spdx/tools-python
+long_description = file: README.md
+long_description_content_type = text/markdown
+classifiers =
+    Intended Audience :: Developers
+    License :: OSI Approved :: Apache Software License
+    Programming Language :: Python :: 2.7
```
pombredanne, line 16: We are dropping Python 2.
```suggestion
    Programming Language :: Python :: 3
```

setup.cfg
```diff
@@ -1,8 +1,49 @@
...
+zip_safe = False
+install_requires =
+    ply
+    rdflib
+    six
+    enum34
+    Click
+    pyyaml
+    xmltodict
+    pyparsing<=1.5.7;python_version<="2.8"
```
pombredanne, line 29: We are dropping Python 2
```suggestion
```

setup.cfg
```diff
@@ -1,8 +1,49 @@
...
+include_package_data = True
+python_requires = >=2.7
```
pombredanne, line 31:
```suggestion
python_requires = >=3.6
```

setup.cfg
```diff
@@ -1,8 +1,49 @@
...
+    tests
+    examples.*
+    tests.*
+
+[options.entry_points]
+console_scripts =
+    pyspdxtools_convertor = spdx.cli_tools.convertor:main
+    pyspdxtools_parser = spdx.cli_tools.parser:main
 
 [bdist_wheel]
```
pombredanne, line 45:
```suggestion
```

setup.cfg
```diff
@@ -1,8 +1,49 @@
...
-universal=1
+universal = 1
```
pombredanne, line 46: We are no longer universal, but Python 3 only
```suggestion
```

setup.py
```diff
@@ -1,61 +0,0 @@
-#!/usr/bin/env python
```
pombredanne, line 1: It is best to keep for now a shell setup.py such as this:
```
#!/usr/bin/env python

import setuptools

if __name__ == "__main__":
    setuptools.setup()
```

pyproject.toml
```diff
@@ -0,0 +1,5 @@
+[build-system]
+requires = ["setuptools>=44", "wheel", "setuptools_scm[toml]>=3.4.3"]
```
pombredanne, line 2: I am not sure we want to switch to `setuptools_scm` just now. Let's have a ticket and do it later, and keep a manual crafted version for now.

pombredanne (dismissed):
Thanks! See my comments inline for you review.

pyproject.toml
pombredanne: Actually this makes sense. Let's do it.

pombredanne:
@KOLANICH can you also add your DCO signoff? This was an informal requirement which was not consistently applied until now... this way I can then merge.

pombredanne:
This has been merged after a rebase.
Thanks!

pombredanne:
Fat fingers!
Not merged yet. Sorry for the noise

armintaenzertng:
@KOLANICH, do you still plan to address these change requests?

@pombredanne, are these changes still up-to-date?

pombredanne:
> @pombredanne, are these changes still up-to-date?

I think they are... but your call to verify this.

armintaenzertng:
Thanks @KOLANICH for updating this! :)

Removing `setup.py` seems like a breaking change, though, and CircleCI workflows also depend on it. I think keeping a minimal `setup.py` as Philippe [suggested here](https://github.com/spdx/tools-python/pull/156/files#r645972104) would be the best option for now.

armintaenzertng:
Thanks for the update, @KOLANICH!
CircleCI is still broken, though. Do you plan to look into this?

armintaenzertng:
@pombredanne, this still has a blocker from you, would you mind to revisit/lift this? :)

pyproject.toml
```diff
@@ -0,0 +1,43 @@
...
+requires-python = ">=3.6"
+dependencies = ["ply", "rdflib", "click", "pyyaml", "xmltodict"]
+dynamic = ["version"]
+
+[project.scripts]
+pyspdxtools_convertor = "spdx.cli_tools.convertor:main"
+pyspdxtools_parser = "spdx.cli_tools.parser:main"
+
+[tool.setuptools]
+zip-safe = false
```
jayvdb, line 33: `zip-safe` isnt important any more (c.f. https://setuptools.pypa.io/en/latest/deprecated/zip_safe.html) , but it doesnt hurt to keep tracking this.
Could you add a comment here that this package isnt zip-safe because of https://github.com/spdx/tools-python/issues/257

pyproject.toml
```diff
@@ -0,0 +1,43 @@
...
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+]
+urls = {Homepage = "https://github.com/spdx/tools-python"}
+requires-python = ">=3.6"
+dependencies = ["ply", "rdflib", "click", "pyyaml", "xmltodict"]
+dynamic = ["version"]
+
+[project.scripts]
+pyspdxtools_convertor = "spdx.cli_tools.convertor:main"
```
jayvdb, line 29: this is going to rename the scripts, which means that the docs need updating. at least the readme https://github.com/spdx/tools-python#command-line-usage

And if we're renaming these, which IMO is a good idea as these names are too generic, can we please change "convertor" to "converter" in order to be using the main word in the dictionaries. c.f. https://thecontentauthority.com/blog/convertor-vs-converter

jayvdb (changes requested):

KOLANICH:
>just noting this is not mentioned at as belonging in project table at https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#setuptools-specific-configuration

That doc mentions only setuptools-specific configuration. The most of `pyproject.toml` is the standardized one for all the major backends except `poetry` according to PEP 621.

>And if we're renaming these, which IMO is a good idea as these names are too generic, can we please change "convertor" to "converter" in order to be using the main word in the dictionaries.

To be honest, `convertor` feels better for me and I guess for people too: searching in Bing (Google doesn't give out this info at least when JS is disabled) reveals:
convertor `566 000 000 results`
converter `564 000 000 results`

The rest of suggestions are implemented, thanks.

pyproject.toml
jayvdb: See https://trends.google.com.au/trends/explore?geo=AU&q=convertor,converter for more evidence that convertor is not common.
And look up the word in dictionaries.

jayvdb: v0.7.0 will be dropping support for Python 2, which is already a breaking change.

And `parser` & `convertor` were added after v0.6.1, so they havent been released yet.  Now is the time to change them.

https://pypi.org/project/spdx-tools/0.6.1/ was released August 2019, before https://github.com/spdx/tools-python/pull/148

nicoweidner: Thanks for pointing it out! I had only seen the date of the Github release (which was created July 2021, two years after the tag and pypi release were created...) and made assumptions without actually checking. Let's rename then.

nicoweidner (dismissed):
Thanks for updating the setup scripts! I won't be able to merge until the appveyor issue is resolved, though...

@pombredanne Anything missing in your opinion? You are still set to "changes requested".

pyproject.toml
```diff
@@ -0,0 +1,49 @@
...
+    "Programming Language :: Python :: 3.7",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+]
+urls = {Homepage = "https://github.com/spdx/tools-python"}
+requires-python = ">=3.6"
+dependencies = ["ply", "rdflib", "click", "pyyaml", "xmltodict"]
+dynamic = ["version"]
```
nicoweidner, line 29: Didn't know about this, pretty cool!
In case anyone else stumbles across this: https://pypi.org/project/setuptools-scm/, look for "Default versioning scheme"

nicoweidner (approved):
Thanks for rebasing again, @KOLANICH! I will go ahead and merge this before new conflicts arise

jayvdb:
Please see https://github.com/spdx/tools-python/pull/362

PR #157, opened 2021-02-24
Added .editorconfig according to PEP 8

KOLANICH:


pombredanne:
@KOLANICH Thanks for this. Do you have some pointer or URL to explain what this about? Could this be instead a local, ignored file?

KOLANICH:
>Do you have some pointer or URL to explain what this about?

https://editorconfig.org/

> Could this be instead a local, ignored file?

Technically it can. But it defeats the purpose.

nicoweidner:
@pombredanne I would greatly appreciate it if there were well-defined code style settings checked in. I would even like to integrate a corresponding check in the CI.

As to why:
- To avoid seeing ~50 warnings in every file in the IDE, as is currently the case
- To avoid contributors reformatting code with different settings in each PR because there is no defined standard for the repo

jayvdb (approved):

nicoweidner (approved):
Thanks, and sorry for taking so long to get around to this!

KOLANICH:
No problem, it's Open Source where noone owes anyone anything :)

Issue #158, opened 2021-04-23
Some tests fail on Python 3.9 with `TypeError: __init__() got an unexpected keyword argument 'encoding'`


cole-h:
```
_________________________ TestParser.test_yaml_parser __________________________

self = <tests.test_jsonyamlxml_parser.TestParser testMethod=test_yaml_parser>

    def test_yaml_parser(self):
        parser = yamlparser.Parser(Builder(), StandardLogger())
        test_file = utils_test.get_test_loc('formats/SPDXYamlExample.yaml')
        with io.open(test_file, encoding='utf-8') as f:
            document, _ = parser.parse(f)
        expected_loc = utils_test.get_test_loc('doc_parse/expected.json')
>       self.check_document(document, expected_loc)

tests/test_jsonyamlxml_parser.py:59:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/test_jsonyamlxml_parser.py:41: in check_document
    expected = json.load(ex, encoding='utf-8', object_pairs_hook=OrderedDict)
/nix/store/y1mhcjrx951lbdbf69v2wng33gc7lx1j-python3-3.9.2/lib/python3.9/json/__init__.py:293: in load
    return loads(fp.read(),

[snip]

E       TypeError: __init__() got an unexpected keyword argument 'encoding'

/nix/store/y1mhcjrx951lbdbf69v2wng33gc7lx1j-python3-3.9.2/lib/python3.9/json/__init__.py:359: TypeError

[snip]

FAILED tests/test_jsonyamlxml_parser.py::TestParser::test_json_parser - TypeE...
FAILED tests/test_jsonyamlxml_parser.py::TestParser::test_xml_parser - TypeEr...
FAILED tests/test_jsonyamlxml_parser.py::TestParser::test_yaml_parser - TypeE...
```

All is well in Python 3.8.

PR #159, opened 2021-04-23
test_jsonyamlxml_parser: remove deprecated json.load encoding argument

cole-h:
It was deprecated in Python 3.1 and removed in Python 3.9.

Fixes https://github.com/spdx/tools-python/issues/158.


pombredanne:
Thanks. LGTM. The failures in CI are transient as @licquia has pending PRs to fix this!

PR #162, opened 2021-06-04
Tie click version for Python 2.x.

### LINKED ISSUE #161: CI failures: click has dropped support for Python 2.7
From the release notes for click (https://click.palletsprojects.com/en/8.0.x/changes/#version-8-0-0):

* Drop support for Python 2 and 3.5.

This means that all Python 2.7 tests fail in CI.

We should tie the click version to the Python version, so click 7.x is forced for Python 2.7.

licquia:
Click dropped support for all 2.x versions and all 3.5 or earlier
versions of Python.  So, in order to support those versions, we
must tie the click version to click 7.x.

Since we're still actively testing these older Python versions
in CI, this should fix a number of issues in CI.

Fixes #161.


licquia:
Per comment from @pombredanne in PR #164, it's probably better to just drop testing for 2.7, 3.4, and 3.5.  Adjusted accordingly.

pombredanne (approved):
Looking good!

pombredanne:
I am merging anyway as this PR is to fix the CI.
@licquia FYI I enabled the DCO bot on this repo.

PR #168, opened 2021-06-08
Support multiple packages per SPDX document.

### LINKED ISSUE #79: Add support for parsing more than one package
The current implementation only supports parsing a single package in the SPDX document. There can be one or more packages as mentioned [here](https://spdx.org/sites/cpstandard/files/pages/files/spdxversion2.1.pdf#page=18). We should add support for parsing more than one package.

licquia:
This set of patches adds support for multiple packages per document, by adding document.packages as an array.  The API for managing multiple packages is set up to mimic the other APIs for managing multiple objects (files, external licenses, etc.).

To preserve backwards compatibility, this patch introduces document.package as a property; this property is implemented using document.packages behind the scenes, and triggers a DeprecationWarning.

Support is included for parsing documents with multiple packages, as well as writing such documents.

Fixes #79.

### DIFF
--- a/spdx/document.py
+++ b/spdx/document.py
@@ -13,6 +13,8 @@ from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import unicode_literals
 
+import warnings
+
 from functools import total_ordering
 
 from spdx import config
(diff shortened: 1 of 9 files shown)


pombredanne (commented):
LTGM... I am not too worried about the CI hiccups... but let me see how the warning behaves in practice

pombredanne:
All good to me. Merging. Thanks!

PR #170, opened 2021-06-15
Add a sbom_generator command-line tool.

licquia:
This tool generates a SPDX software bill of materials (SBOM)
from a Python project.  It's intended to make it easy to generate
SBOMs from Python projects with a simple dev dependency and a single
command.


pombredanne (changes requested):
@licquia Thanks! that's great but IMHO:
- this would be best in the examples directory or a separate repo especially since this is making outgoing networks calls on a whole current installation. That would need to be documented too.  
- This would benefit from some doc and tests
- This introduces an undeclared dependency with `importlib_metadata`

pombredanne:
Gentle ping.

licquia:
Yes, sorry for not following up.  Since development on this and on the original tool have both moved on, I'm going to submit a new PR for this.

PR #173, opened 2021-06-22
introduce message handler

tardyp:
There are a lot of 

```
            messages = messages + ["ExternalPackageRef has no category."]
```

in this codebase..

This is not very pythonic, and does not contain much context information, which makes it difficult to understand what part of the 8MB spdx dump file generated by proprietary tool is not recognised.

This is a first simple code to start discussion on the problematic.
Will integrate in the rest of the code when we agree on the API.

### DIFF
--- a/spdx/document.py
+++ b/spdx/document.py
@@ -53,29 +53,21 @@ class ExternalDocumentRef(object):
         Validate all fields of the ExternalDocumentRef class and update the
         messages list with user friendly error messages for display.
         """
-        messages = self.validate_ext_doc_id(messages)
-        messages = self.validate_spdx_doc_uri(messages)
-        messages = self.validate_checksum(messages)
-
-        return messages
+        self.validate_ext_doc_id(messages)
+        self.validate_spdx_doc_uri(messages)
+        self.validate_checksum(messages)
 
     def validate_ext_doc_id(self, messages):
         if not self.external_document_id:
-            messages = messages + ["ExternalDocumentRef has no External Document ID."]
-
-        return messages
+            messages.append("ExternalDocumentRef has no External Document ID.")
 
     def validate_spdx_doc_uri(self, messages):
         if not self.spdx_document_uri:
-            messages = messages + ["ExternalDocumentRef has no SPDX Document URI."]
-
-        return messages
+            messages.append("ExternalDocumentRef has no SPDX Document URI.")
 
     def validate_checksum(self, messages):
         if not self.check_sum:
-            messages = messages + ["ExternalDocumentRef has no Checksum."]
-
-        return messages
+            messages.append("ExternalDocumentRef has no Checksum.")
 
 
 def _add_parens(required, text):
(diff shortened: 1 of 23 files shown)


tardyp:
hi @pombredanne  I cannot find a open source oss managment tool were you are not involved :)

spdx/document.py
```diff
@@ -366,58 +358,52 @@ def validate(self, messages):
...
-        messages = self.validate_ext_document_references(messages)
-        messages = self.validate_creation_info(messages)
-        messages = self.validate_packages(messages)
-        messages = self.validate_extracted_licenses(messages)
-        messages = self.validate_reviews(messages)
-        messages = self.validate_snippet(messages)
-        # messages = self.validate_annotations(messages)
-        # messages = self.validate_relationships(messages)
-
+        if isinstance(messages, list):
```
pombredanne, line 361: Here you are overriding entirely the messages argument, right? What does this mean then? What's the impact?

spdx/document.py
```diff
@@ -366,58 +358,52 @@ def validate(self, messages):
...
+        self.validate_spdx_id(messages)
+        self.validate_namespace(messages)
+        self.validate_ext_document_references(messages)
+        self.validate_creation_info(messages)
+        self.validate_packages(messages)
+        self.validate_extracted_licenses(messages)
+        self.validate_reviews(messages)
+        self.validate_snippet(messages)
+        # self.validate_annotations(messages)
+        # self.validate_relationships(messages)
```
pombredanne, line 377: Let's get rid of dead commented code

spdx/relationship.py
```diff
@@ -101,18 +102,15 @@ def validate(self, messages):
...
         r_type = self.relationship.split(" ")[1]
         if r_type not in [name for name, _ in RelationshipType.__members__.items()]:
-            messages = messages + [
+            messages.append(
                 "Relationship type must be one of the constants defined in "
                 "class spdx.relationship.Relationship"
-            ]
-            return messages
+            )
+            return False
```
pombredanne, line 114: Could we instead always return ErrorMessages? here and elsewhere

pombredanne (changes requested):
Thanks! See some comments for your consideration.

spdx/document.py
tardyp: I think this was in order to keep API compatibility. but this may not make much sense as there might not be that many users we want to avoid braking.

I'll change that to an explicit error

spdx/document.py
tardyp: well, I think it does not make much sense to have it commented out. was introduced here: https://github.com/spdx/tools-python/commit/209a92563ea321403e0e8abc2308644367d73996

spdx/relationship.py
tardyp: The docstring is different for that task, we are supposed to return True if the relationship is valid. I'll fix and change docstrings

pombredanne (approved):
LGTM.... Thank you!

PR #174, opened 2021-06-25
Factorise parsing and conversion

tardyp:
The example dir looks very confusing with lots of similar code, and the CLI is actually using it.

This PR provide factorization, so that it is easier to write a custom converter without having to copy paste too much boileplate.

The tests show that the conversion is not fully stable.

fixing that is out of scope of this PR, and requires more investigation


tardyp:
I did not realize this is not using pytest. I would suggest to use it as it is defacto standard, and has a lot of nice functionality to improve testability

pombredanne (approved):
LGTM! Thank you ++

PR #175, opened 2021-06-28
python3.9 is only present on 2019 images which is not the default

tardyp:
https://www.appveyor.com/docs/build-environment/ : says 2015 is the default

https://www.appveyor.com/docs/windows-images-software/ : says 3.9 is only on 2019


pombredanne (approved):
Thanks! LGTM :+1:

PR #178, opened 2021-06-30
Use short version id for job names

pombredanne:
Signed-off-by: Philippe Ombredanne <pombredanne@nexb.com>


pombredanne:
This no longer applies.

Issue #177, opened 2021-06-28
use pytest


tardyp:
This codebase is using standard python unittest framework

pytest is arguably more advanced and defacto standard for concise unit testing.

- fixture based testing (tempfile, etc)
- module function based testing without the need to define class
- pytest.parametrized is also very useful (to me)

I would propose to include pytest as a dev dependency, so that it is easier to write tests for spdx/tools-python

We can keep the tests as they are right now, we don't need to translate them as pytest also support standard test classes.

I'll do the PR, if maintainers are agreeing to the change.

Issue #176, opened 2021-06-28
drop python < 3.6 support


tardyp:
there are a bunch of future import and six code in this codebase that is not useful.

I propose that we drop them to cleanup the code

PR #179, opened 2021-07-02
Pytest + py2.7 support removal + CI fixes

tardyp:
Fix: #177 
Fix: #176 

There is a lot of search replace here, I separated each in separate commit for easier review.


pombredanne (approved):
Thanks! This is a lot of tedious work and done in a pristine and clean way!
Merging now.

PR #180, opened 2021-07-05
Draft: bug fixing

yanyag:
Support writing and parsing "filesAnalyzed" into XML/JSON/YAML. 
Fix parsers to support multiple packages per document.
All the tests are passing.

### DIFF
--- a/spdx/parsers/jsonyamlxml.py
+++ b/spdx/parsers/jsonyamlxml.py
@@ -1046,6 +1046,9 @@ class PackageParser(BaseParser):
         - package: Python dict with Package Information fields in it
         """
         if isinstance(package, dict):
+            # Avoid CardinalityError on the next package
+            self.builder.reset_package()
+
             self.parse_pkg_name(package.get("name"))
             self.parse_pkg_id(package.get("SPDXID"))
             self.parse_pkg_version(package.get("versionInfo"))
@@ -1053,6 +1056,7 @@ class PackageParser(BaseParser):
             self.parse_pkg_supplier(package.get("supplier"))
             self.parse_pkg_originator(package.get("originator"))
             self.parse_pkg_down_location(package.get("downloadLocation"))
+            self.parse_pkg_files_analyzed(package.get("filesAnalyzed"))
             self.parse_pkg_verif_code_field(package.get("packageVerificationCode"))
             self.parse_pkg_homepage(package.get("homepage"))
             self.parse_pkg_source_info(package.get("sourceInfo"))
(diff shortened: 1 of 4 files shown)


spdx/parsers/jsonyamlxml.py
```diff
@@ -1181,11 +1185,31 @@ def parse_pkg_down_location(self, pkg_down_location):
...
 
+    def parse_pkg_files_analyzed(self, pkg_files_analyzed):
+        """
+        Parse Package files analyzed
+        - pkg_files_analyzed: Python boolean or str/unicode
+        """
+        # Files Analyzed optional
+        if pkg_files_analyzed is None:
+            return
+        if isinstance(pkg_files_analyzed, six.string_types) or isinstance(pkg_files_analyzed, bool):
```
tardyp, line 1196: according to the spec, https://spdx.github.io/spdx-spec/3-package-information/
filesAnalysed can only be a boolean.

So I would have stored the conversion to bool.

Also, we are removing six in #179 , so please don't use six for new code/

spdx/parsers/jsonyamlxml.py
```diff
@@ -1181,11 +1185,31 @@ def parse_pkg_down_location(self, pkg_down_location):
...
+                self.more_than_one_error("PKG_FILES_ANALYZED")
+        else:
+            self.value_error("PKG_FILES_ANALYZED", pkg_files_analyzed)
+
     def parse_pkg_verif_code_field(self, pkg_verif_code_field):
         """
         Parse Package verification code dict
         - pkg_verif_code_field: Python dict('value':str/unicode, 'excludedFilesNames':list)
         """
+        if self.document.packages[-1].files_analyzed == False:
```
tardyp, line 1211:
```suggestion
        if not self.document.packages[-1].files_analyzed:
```

tardyp: why testing on packages[-1]? Sounds like a hack assuming the current package being parser is the last one.

I think we may rather want to do this test in the caller function

spdx/parsers/tagvalue.py
```diff
@@ -1183,6 +1183,7 @@ def p_package_name(self, p):
                 value = p[2].decode(encoding="utf-8")
             else:
                 value = p[2]
+            self.builder.reset_package()
```
tardyp, line 1186: not sure why this change. Can you please explain?

spdx/writers/jsonyamlxml.py
```diff
@@ -116,16 +116,20 @@ def create_package_info(self, package):
         package_object["SPDXID"] = self.spdx_id(package.spdx_id)
         package_object["name"] = package.name
         package_object["downloadLocation"] = package.download_location.__str__()
-        package_object["packageVerificationCode"] = self.package_verification_code(
-            package
-        )
+        if package.files_analyzed in [True, None]:
```
tardyp, line 119: as the default value for this field is True, we should initialize the field to True.

this test is really not obvious, and will lead to bugs in the future

spdx/writers/jsonyamlxml.py
```diff
@@ -116,16 +116,20 @@ def create_package_info(self, package):
...
+                package
+            )
         package_object["licenseConcluded"] = self.license(package.conc_lics)
         package_object["licenseInfoFromFiles"] = list(
             map(self.license, package.licenses_from_files)
         )
         package_object["licenseDeclared"] = self.license(package.license_declared)
         package_object["copyrightText"] = package.cr_text.__str__()
 
+        if package.has_optional_field("files_analyzed"):
```
tardyp, line 130: as per spec: 3.8.3 Cardinality: Optional, one. If omitted, the default value of true is assumed.

so imho we should really initialize this to True, an only write if it is set to False

```suggestion
        if package.files_analyzed:
```

tardyp (changes requested):

tardyp:
hi @yanyag can you please rebase this PR?

now that the six removal PR is merged, there are a few conflicts.

tardyp:
hi @yanyag  @itraviv ,

Please remember you still have this PR open that needs rebasing. It still contains merge conflicts.

Now the PR contains more fixes than initially pushed, and it becoming less trivial.
I think it will need more unit tests, with 2.2 file examples which proves that your code is actually working.

pombredanne:
Gentle ping :)

itraviv:
hi, we will probably need some time until we'll be able to have this MR ready due to capacity. will ping when ready.

pombredanne:
gentle ping :) Any update?

armintaenzertng:
As this PR is still in draft mode and has not been receiving vital updates for over a year, I suggest closing it.

@yanyag, any objections?

nicoweidner:
Closing this due to inactivity, please ping if it should be reopened.

Issue #182, opened 2021-07-16
Package requires SHA1 hash but spec does not require it


njv299:
The SPDX spec for the 'Package Checksum' field does not specify that it _must_ be SHA1, but the current Package class validation logic requires it to be.

This is the relevant part of the spec:  
![image](https://user-images.githubusercontent.com/210219/125990315-bfa8e2a6-a1b5-400d-ab24-d6eecfe53008.png)

And this is the section of the code which is doing improper validation (from package.py):
![image](https://user-images.githubusercontent.com/210219/125990452-1c86fe4f-fa58-49e3-b281-53025f38d837.png)

In my opinion the spec itself should also be relaxed with respect to File Checksum requiring SHA1, but that is an argument for a different venue. I'm guessing that this bug is due to copy-paste from the File class, which rightfully checks for a SHA1 hash.

njv299:
I created [this PR](https://github.com/spdx/tools-python/pull/183) to remove the mentioned code.

pombredanne:
All fixed, merged and released. 
Thanks!

PR #183, opened 2021-07-16
Removed check that packages have SHA1 checksum, per spec

njv299:
This is a simple attempt to fix this issue:

https://github.com/spdx/tools-python/issues/182

Essentially, it just removes the check for a Package checksum to be SHA1 since the SPDX spec does not require that.


tardyp (approved):

pombredanne (changes requested):
Thank you ++ 
Do you mind to add a test with this?

njv299:
@pombredanne I added a test as requested. 

Thanks!

pombredanne (approved):
LGTM. Thank you ++

pombredanne:
Some of the tests on Circle and Appveyor take forever to complete.

tardyp:
@pombredanne they don't.. they will never be done, because they are not done for PRs.
you should remove "continuous-integration/appveyor/branch" from the required tests in the protected branch area.
This test is only done *after* merge.

pombredanne:
@tardyp thanks! You rock. I fixed this.

Issue #186, opened 2021-07-22
Incomplete FileType enumeration


njv299:
The current allowed values for FileType are incomplete relative to what is listed in the spec.

The values supported by the current repo are:
![image](https://user-images.githubusercontent.com/210219/126668206-a84b00dd-231c-400d-9664-403be930667d.png)
(https://github.com/spdx/tools-python/blob/master/spdx/file.py#L20)

The values allowed by the spec are:
![image](https://user-images.githubusercontent.com/210219/126668114-fb78cae2-5c69-47ed-92c0-36faa9971b03.png)

njv299:
PR for this is here: https://github.com/spdx/tools-python/pull/187

PR #187, opened 2021-07-22
Added missing FileType values allowed by spec

njv299:
Adds FileType values allowed by spec but not currently listed in FileType class. 

Resolves [issue #186](https://github.com/spdx/tools-python/issues/186) 

Signed-off-by: Nathan Voss <nathan@finitestate.io>


spdx/file.py
```diff
@@ -22,6 +22,13 @@ class FileType(object):
     BINARY = 2
     ARCHIVE = 3
     OTHER = 4
+    APPLICATION = 5
```
tardyp, line 25: looks compliant to the spec:

https://spdx.github.io/spdx-spec/4-file-information/

```
SOURCE if the file is human readable source code (.c, .html, etc.);
BINARY if the file is a compiled object, target image or binary executable (.o, .a, etc.);
ARCHIVE if the file represents an archive (.tar, .jar, etc.);
APPLICATION if the file is associated with a specific application type (MIME type of application/*);
AUDIO if the file is associated with an audio file (MIME type of audio/* , e.g. .mp3);
IMAGE if the file is associated with an picture image file (MIME type of image/*, e.g., .jpg, .gif);
TEXT if the file is human readable text file (MIME type of text/*);
VIDEO if the file is associated with a video file type (MIME type of video/*);
DOCUMENTATION if the file serves as documentation;
SPDX if the file is an SPDX document;
OTHER if the file doesn't fit into the above categories (generated artifacts, data files, etc.)
```

tardyp (approved):
looks good.

As we are now python3 only, we could migrate this class to enum to get better repr()

we can even use enum's auto
https://docs.python.org/3/library/enum.html#enum.auto

pombredanne (approved):
LGTM! Thanks .. merging

PR #191, opened 2021-09-20
Add a SBOM command-line generator tool.

### LINKED ISSUE #171: Add sbom_generator command-line tool
We've developed a simple tool for generating a software bill of materials from a Python project, intended to be quick and simple.  I've submitted it as PR #170.

licquia:
This patch add the sbom_generator utility, which examines a Python
project and outputs a SPDX SBOM to standard output.

Fixes #171.

Signed-off-by: Jeff Licquia <licquia@linuxfoundation.org>

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,13 @@ Ex. : `convertor -f rdf data/SPDXRdfExample.xyz -o output.xml`
 * for help - use `convertor --help`
 
 
+3. **SBOM GENERATOR** (for generating a software bill of materials):
+* Use   `sbom_generator <project>`   where  `<project>`  is the name of the top-level project.
+(Note: If using a version of Python older than 3.8, you will need to install `importlib_metadata` to use `sbom_generator`.)
+Try running :   `sbom_generator spdx-tools`.
+
+* for help - use `parser --help`
+
 
 # Installation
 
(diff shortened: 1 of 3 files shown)


licquia:
Addressing feedback from the previous PR (#170):

- I've improved the docs to be in line with the other CLI tools.  It doesn't appear any of the tools have tests, which I agree we should have.
- I've also added a note to the docs describing the importlib_metadata dependency; it is only needed for Python 3.7 or earlier.

I'll leave the question of whether we need this as a CLI tool for others to discuss; suffice it to say that there is demand for this kind of tool.

pombredanne (changes requested):
@licquia thanks for the udpate!
In https://github.com/spdx/tools-python/pull/170#pullrequestreview-684158959 I wrote:

>  this would be best in the examples directory or a separate repo especially since this is making outgoing networks calls on a whole current installation. 
> This would benefit from some doc and tests
> This introduces an undeclared dependency with importlib_metadata

Making network calls in the core SPDX parsing library is not something that is going to be acceptable for corporate users that run in an isolated network. There is also no tests. Do you mind to move this to the `examples/` directory instead?

maxhbr:
Hey, I am not sure if we would want to have that within the tools. I would assume that we should focus on the library first. Especially since that is in the meantime solved by https://github.com/nexB/python-inspector

PR #192, opened 2021-10-12
Fix output handling with filesAnalyzed

lhh:


lhh:
This needs more work; I'll reopen when it's passing.

Issue #193, opened 2021-10-13
Fix output handling with filesAnalyzed


pombredanne:
This is a follow up to the closed PR https://github.com/spdx/tools-python/pull/192 by @lhh so we can have a discussion on the issue here.

The two commits messages were:
- Include filesAnalyzed in machine-readable output 82be1b3

- Fix filesAnalyzed handling f2a13d6
When filesAnalyzed is False, several sections should be omitted
when omitting rather than being presented as empty sets or "None"
SPDX 2.2, sec:
  3.8.3 - files analyzed specification
  3.9.3 - verification code
  3.14.3 - licenses from files

armintaenzertng:
Closing this issue as it seems the discussion is done by now. Please speak up if it should be reopened.

PR #195, opened 2021-10-19
Fix up a few filesAnalyzed issues

lhh:


lhh:
Looks like circleci stuff on mac is broken. I'm not quite sure how to fix it.

pombredanne:
And sorry if I did update you branch by mistake with the latest. Let's pretend it was because of fat fingers.

lhh:
Yes, of course. I'll add some tests.

pombredanne:
@lhh Thank you! you rock.

dholth:
Should `package_object["packageVerificationCode"] = self.package_verification_code(` be optional or 'mandatory missing' per spec version 2.2 https://spdx.github.io/spdx-spec/package-information/#79-package-verification-code-field

pombredanne:
> Should `package_object["packageVerificationCode"] = self.package_verification_code(` be optional or 'mandatory missing' per spec version 2.2 https://spdx.github.io/spdx-spec/package-information/#79-package-verification-code-field

@dholth this code is a source of troubles more often than not. We should make it optional in this library IMHO.

armintaenzertng:
@lhh, do you still plan on updating this PR?

lhh:
Ah, yes - my apologies. I'll address the above issues shortly.

jasinner (dismissed):
LGTM. I verified this fixed the bug described in #193 


@goneall  please merge this.

pombredanne (approved):

tests/utils_test.py
```diff
@@ -106,7 +106,11 @@ def sort_nested(data):
         new_data = {}
         for k, v in data.items():
             if isinstance(v, list):
-                v = sorted(v)
+                try:
+                    v = sorted(v)
+                except TypeError:
+                    # print(f'Warning: Cannot sort {v}')
```
nicoweidner, line 112: nit: commented out code. Printing a warning seems reasonable here, so may I ask why it was commented out?

tests/utils_test.py
```diff
@@ -106,7 +106,11 @@ def sort_nested(data):
         new_data = {}
         for k, v in data.items():
             if isinstance(v, list):
-                v = sorted(v)
+                try:
+                    v = sorted(v)
```
nicoweidner, line 110: This is the case where `data` is a dict and `data[k]` is of type `list`. Why isn't `sort_nested` called recursively here? In the situation where `data` is a list and a given value `v` is of type `list`, `sort_nested` is called on it.

nicoweidner (commented):
I left two minor questions. Thanks for the contribution!

tests/utils_test.py
lhh: You're correct, it's appropriate to call sort_nested() here.

tests/utils_test.py
lhh: I used this for debugging. This was printing out dictionaries which pop up since calling "sorted()" builtin is undefined behavior when one or more of the entries in the a list is a dict (or, in the case of this code path, OrderedDict). It's not actually a bug; the SPDX format allows this.

I'll remove the dead code.

tests/utils_test.py
nicoweidner: Cool, thanks for coming back so quickly!

nicoweidner (approved):
Thanks for addressing my questions! I didn't look at the remaining code in detail, but two other reviewes already approved, so this should be good to merge

PR #197, opened 2021-11-12
changes necessary to validate against SPDX jsonschema

### LINKED ISSUE #184: JSON is not copatible with v2.2 spec - "packages" and "files" lists should be outside "documentDescribes"
I see a major gap between the the v2.2 spec and the current python-tool implementation of the JSON format.
In https://github.com/Yash-Varshney/tools-python/issues/9 discussion with @Yash-Varshney, @goneall have processed to add spec issue around it.

<html><body>

Python-tool (presently) | Java-tool (v2.2) | proposed | Spec
-- | -- | -- | --
[MAJOR CHANGE] documentDescribes contains Package and Files | documentDescribes is only a list and packages and files another separate entity | Add spec issue | documentDescribes should only be the top-level package/file that the  document describes, not the entire list - may need more research
Package is a dict | Package is a list in which there is dict ex. [{}] | Add spec issue | This is related to the documentDescribes difference

</body>
</html>

The v2.2 spec points to https://github.com/spdx/spdx-spec/blob/master/schemas/spdx-schema.json schema. In addition, all the other spdx tools implementations are aligned with the spec (java-tools, golang, etc).

I think that we should align the python-tool as well, to complete the v2.2 support (formally declared as WIP now).
Do you guys agree? 
Do you have any plans/roadmap of closing the v2.2 support gap?

Thanks,
Yan

dholth:
These changes were necessary to validate against the SPDX jsonschema https://github.com/spdx/spdx-spec/blob/development/v2.2.2/schemas/spdx-schema.json and to include multiple checksums.

This is a work in progress. No attention has been given to parsing or non-json/yaml formats.

Let me know if this approach is acceptable for the project.

Fixes #184

### DIFF
--- a/spdx/parsers/jsonyamlxml.py
+++ b/spdx/parsers/jsonyamlxml.py
@@ -1486,6 +1486,39 @@ class PackageParser(BaseParser):
             self.value_error("PKG_CHECKSUM", pkg_chksum)
 
 
+def unflatten_document(document):
+    """
+    Inverse of spdx.writers.jsonyamlxml.flatten_document
+    """
+    files_by_id = {}
+    if "files" in document:
+        for f in document.pop("files"):
+            f["name"] = f.pop("fileName")
+            # XXX must downstream rely on "sha1" property?
+            for checksum in f["checksums"]:
+                if checksum["algorithm"] == "SHA1":
+                    f["sha1"] = checksum["checksumValue"]
+                    break
+            if "licenseInfoInFiles" in f:
+                f["licenseInfoFromFiles"] = f.pop("licenseInfoInFiles")
+            files_by_id[f["SPDXID"]] = f
+
+    packages = document.pop("packages")
+    for package in packages:
+        if "hasFiles" in package:
+            package["files"] = [{
+                "File": files_by_id[spdxid]} for spdxid in package["hasFiles"]
+            ]
+        # XXX must downstream rely on "sha1" property?
+        for checksum in package.get("checksums", []):
+            if checksum["algorithm"] == "SHA1":
+                package["sha1"] = checksum["checksumValue"]
+                break
+
+    document["documentDescribes"] = [{ "Package": package} for package in packages ]
+
+    return document
+
 class Parser(
     CreationInfoParser,
     ExternalDocumentRefsParser,
@@ -1503,7 +1536,7 @@ class Parser(
     def json_yaml_set_document(self, data):
         # we could verify that the spdxVersion >= 2.2, but we try to be resilient in parsing
         if data.get("spdxVersion"):
-            self.document_object = data
+            self.document_object = unflatten_document(data)
             return
         self.document_object = data.get("Document")
 
--- a/spdx/creationinfo.py
+++ b/spdx/creationinfo.py
@@ -44,7 +44,7 @@ class Organization(Creator):
     - email: Org's email address. Optional. Type: str.
     """
 
-    def __init__(self, name, email):
+    def __init__(self, name, email=None):
         super(Organization, self).__init__(name)
         self.email = email
 
@@ -80,7 +80,7 @@ class Person(Creator):
     - email: person's email address. Optional. Type: str.
     """
 
-    def __init__(self, name, email):
+    def __init__(self, name, email=None):
         super(Person, self).__init__(name)
         self.email = email
 
--- a/spdx/file.py
+++ b/spdx/file.py
@@ -63,7 +63,7 @@ class File(object):
         self.spdx_id = spdx_id
         self.comment = None
         self.type = None
-        self.chk_sum = chk_sum
+        self.checksums = [None]
         self.conc_lics = None
         self.licenses_in_file = []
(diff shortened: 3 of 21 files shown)


pombredanne:
Thank you ++ This works. ... Great and honored to see you drop by!

dholth:
@pombredanne OK to add Python 3 type annotations?

dholth:
@pombredanne I think this could be merged. It shoud be able to read/write json that is valid according to the spdx jsonschema, and it should be able to write multiple checksums.

Still open is to read multiple checksums, but that further complicates the code. It would also be nice to add a test that validates the output against the spdx spec using the jsonschema Python package.

pombredanne:
> @pombredanne OK to add Python 3 type annotations? OK to format with black?

I am not too keen but OK wrt. Python 3 type annotations and yes black is fine, but it would be best in a completely separate PR as this would create a "wall" in the history

pombredanne:
> I think this could be merged. It shoud be able to read/write json that is valid according to the spdx jsonschema, and it should be able to write multiple checksums.
> 
> Still open is to read multiple checksums, but that further complicates the code. It would also be nice to add a test that validates the output against the spdx spec using the jsonschema Python package.

Thanks! I will need a few days before I can review.

pombredanne:
I am not sure why the macOS tests fail... but Circle CI is a bit flaky at times

RodneyRichardson:
macos tests are failing due to using a now-unsupported version of xcode: See #202 for a fix

nettrino:
@pombredanne let us know if you require any updates to this - it would be useful to get it merged! Thanks

goneall:
@dholth can you rebase this PR to the latest in the main branch?  This should fix the CircleCI errors.

nicoweidner:
CircleCI should be fixed once https://github.com/spdx/tools-python/pull/226 is merged and this PR is rebased on top of the changes.

nicoweidner:
@dholth Will you have time to rebase this PR on `main`? I'd also be happy to do so if you are short on time.

spdx/parsers/jsonyamlxml.py
```diff
@@ -1486,6 +1486,39 @@ def parse_pkg_chksum(self, pkg_chksum):
...
+            package["files"] = [{
+                "File": files_by_id[spdxid]} for spdxid in package["hasFiles"]
+            ]
+        # XXX must downstream rely on "sha1" property?
+        for checksum in package.get("checksums", []):
+            if checksum["algorithm"] == "SHA1":
+                package["sha1"] = checksum["checksumValue"]
+                break
+
+    document["documentDescribes"] = [{ "Package": package} for package in packages ]
```
nicoweidner, line 1518: This does not comply with the json schema, which just has an array of spdx ids here: https://github.com/spdx/spdx-spec/blob/development/v2.3.1/schemas/spdx-schema.json#L219-L226

Even if it's just a temporary internal data structure, I find this highly confusing

nicoweidner (changes requested):
I don't think adding full package objects to `documentDescribes` is a good idea, since it contradicts the spec

meretp:
@dholth As you already pointed out this PR doesn't handle parsing or other file formats. Do you plan on working on these things more or are you okay with me working on them?

spdx/parsers/jsonyamlxml.py
nicoweidner: I don't see the point of flattening the structure in this case though. The only instance where the value is accessed is [this part where the packages are parsed](https://github.com/spdx/tools-python/blob/76272c47035475126bf3252a2d72d100eff372d7/spdx/parsers/jsonyamlxml.py#L1688-L1689), and this can be done just as well without modifying the data structure first. In fact, it [is already done this way](https://github.com/spdx/tools-python/blob/76272c47035475126bf3252a2d72d100eff372d7/spdx/parsers/jsonyamlxml.py#L1704-L1705) (although the fact that this logic is technically duplicated should be fixed).

If resolving links becomes annoying, I would find it more natural to add a utility method that handles it instead of introducing an additional intermediate data structure. This will come with a performance impact if called many times, but I would be surprised if this was an issue.

Additional remark: The current structure only considers files that are included in packages. To my understanding, files should also be allowed to be independent of any package (this might be related to the python tools still being out of date though, it seems like snippets are not considered at all, for example).

nicoweidner: One additional comment: My understanding so far is that in a json document, any `SPDX-Document DESCRIBES SPDX-Package` relationship would be omitted and instead the package id would appear in `documentDescribes` (similar for files and snippets). This would imply that during parsing, we have to create new relationships for elements of `documentDesccribes`. I don't see that happening anywhere.

Relevant discussion: https://github.com/spdx/tools-python/issues/242. Let's see if I'm wrong here :sweat_smile:

dholth:
I would be happy for anyone to pick this up, thanks!

meretp:
Thanks for your work @dholth! I openend another PR which continues this one and will close this one.

PR #209, opened 2022-04-04
Updated README.md

HARDIK-TSH1392:
Corrected the documentation in README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Python library to to parse, validate and create SPDX documents
+# Python library to parse, validate and create SPDX documents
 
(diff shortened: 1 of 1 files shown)


seabass-labrax:
Thanks for making this contribution, @HARDIK-TSH1392! One thing we need before this can be merged is your 'DCO signoff'. This means that you agree to the [Developer Certificate of Origin](https://developercertificate.org/), which is how you give permission to SPDX for your contribution to be released under the appropriate open source license. In this case, it wouldn't matter too much, because the change is probably not copyrightable, but it helps us to keep the record clean!

PR #223, opened 2022-07-28
Reimplmented parse_anything.py

linynjosh:


tests/test_parse_anything.py
```diff
@@ -15,7 +15,7 @@
 from spdx.parsers import parse_anything
 
 
-dirname = os.path.join(os.path.dirname(__file__), "data", "formats")
```
licquia, line 18: Instead of replacing the old tests with the new ones, can we do both sets of tests?

licquia (changes requested):

tests/test_parse_anything.py
linynjosh: Hi Jeff, I added the formats test set back into test_parse_anything. Please tell me if there any other changes you would like me to make.

nicoweidner:
@linynjosh Does this PR fix a specific issue? If it doesn't, can you add a general description of the purpose of the change to the PR?

spdx/parsers/parse_anything.py
```diff
@@ -22,33 +22,25 @@
...
+                p.build()
+            with open(fn) as f:
+                if read_data:
+                    data = f.read()
+                    return p.parse(data)
+                else:
+                    return p.parse(f)
+        except:
+            if i == len(buildermodules) - 1:
+                raise FileTypeError("FileType Not Supported" + str(fn))
```
nicoweidner, line 44: One more comment about this `except:` block: It will hide any exceptions that may be thrown by the parsers. This does not include validation errors (since those do not actually raise exceptions), but if anything goes wrong for other reasons, it won't be visible.

armintaenzertng:
@linynjosh, do you plan to address this PR again?

nicoweidner:
Closing this due to inactivity. Please ping me in case it should be reopened.

Issue #152, opened 2020-10-24
Convertor of cli_tool not working


RishabhBhatnagar:
This is a follow-up of https://github.com/spdx/tools-python/pull/148#discussion_r509255423 .

> I've tried running the convertor. But couldn't run it successfully for any of the ways defined in the README.md file.
>
> The following are the ways I tried and the errors they've raised.
>
>     convertor --infile file.rdf --outfile output.json raised following error:
>     Error: Invalid value for '--to' / '-t': invalid choice: undefined. (choose from json, rdf, yaml, xml, tag)
>     convertor --from rdf file.rdf --to json output.json raised following error:
>     spdx.parsers.builderexceptions.FileTypeError: INPUT FILETYPE NOT SUPPORTED. (only RDF and TAG format supported)
>     convertor -f rdf file.rdf -o output.json raised following error:
>     Error: Invalid value for '--to' / '-t': invalid choice: undefined. (choose from json, rdf, yaml, xml, tag)
>
> If you think I might've made some mistake while running the tool, you can check how I ran it in this [google colab file](https://colab.research.google.com/drive/1UUx2grrfII71fwCevYgbtCOted5qkS4t?usp=sharing)

TraceyOnim:
Is there any follow-up on the issue? I'm also facing the same error .

pombredanne:
@TraceyOnim not yet... help would be mucho welcomed!

nicoweidner:
It looks like the issues reported here were actually fixed in https://github.com/spdx/tools-python/pull/174; the errors are gone. However, the same PR caused an issue with the first example where both `infile` and `outfile` is used. That is caught by [this block](https://github.com/spdx/tools-python/blob/3efab86fd735cb527032c959417668bc9a2d3de1/spdx/cli_tools/convertor.py#L79-L81) and just prints the help message.

meretp:
I'll have a look.

nicoweidner:
Fixed via https://github.com/spdx/tools-python/pull/235

PR #235, opened 2022-10-13
[issue-152] fix cli convertor and add testcase

meretp:
fixes issue #152 

I decided to extract a method to get the infile and outfile from the given input arguments to also add a TestCase for this functionality.

### DIFF
--- a/tests/test_cli_convertor.py
+++ b/tests/test_cli_convertor.py
@@ -1,0 +1,86 @@
+# Copyright (c) 2022 spdx tool contributors
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


tests/test_cli_convertor.py
```diff
@@ -0,0 +1,86 @@
...
+        src = ('infile.in', 'outfile.out')
+        from_ = 'rdf'
+        to = 'json'
+
+        infile, outfile = determine_infile_and_outfile(infile_given, outfile_given, src, from_, to)
+
+        infile_path = os.path.splitext(src[0])[0]
+        infile_given = infile_path + "." + from_
+        outfile_path = os.path.splitext(src[1])[0]
+        outfile_given = outfile_path + "." + to
```
nicoweidner, line 46: This is basically a reimplementation of the logic that we are testing. For tests, I would suggest just statically providing the expected result and asserting on it (e.g. including `expected_infile_path = infile.rdf`)

nicoweidner: (the same applies to the other test cases, of course)

nicoweidner (dismissed):
Thanks, really nice improvement for your first PR!
I just left a minor comment regarding the test setup. I don't think this will need another review round, so I will already approve now

nicoweidner (approved):
Thanks for addressing my comments!

PR #236, opened 2022-10-14
[issue-222] fix parsing of tv-files with multiple packages

### LINKED ISSUE #222: TV SBOM with multiple packages not read correctly
I have an SBOM that only contains packages, setting FilesAnalyzed to false this should be according to standard.

Looking at the SBOM examples from SPDX, this example4 causes the same problem when trying to be read by python library
https://github.com/spdx/spdx-examples/blob/master/example4/spdx/example4-bin.spdx

Error output: spdx.parsers.builderexceptions.OrderError: File::SPDXID

Strange as this error should only be set, when there is a file entry

Trying to understand the error, it seems it is calling "p_spdx_id(self, p):" for the next package and as self.builder.doc_spdx_id_set is True and self.builder.package_spdx_id_set is True it is calling self.builder.set_file_spdx_id(self.document, value) causing this error.

Looks like it is not "closing" the package and assumes the next package is a file. If somebody with a deeper understanding of the code has some idea what the problem is, would be much appreciated.

meretp:
fixes #222 

I also added a new testfile so that parsing of TV-files with multiple packages is also tested. The new testfile needs to be skipped in the [test_write_anything.py](https://github.com/spdx/tools-python/blob/main/tests/test_write_anything.py) since conversion of SPDX 2.2 files is not implemented yet and yields to failures of the test.  
This PR also fixes parsing of .spdx files which were parsed as .rdf files before. 

### DIFF
--- a/spdx/parsers/tagvaluebuilders.py
+++ b/spdx/parsers/tagvaluebuilders.py
@@ -602,6 +602,7 @@ class PackageBuilder(object):
         name - any string.
         Raise CardinalityError if package already defined.
         """
+        self.reset_package()
         self.package_set = True
         doc.add_package(package.Package(name=name))
         return True
--- a/tests/data/formats/SPDXSBOMExample.tag
+++ b/tests/data/formats/SPDXSBOMExample.tag
@@ -1,0 +1,64 @@
+# Document Information
+
+SPDXVersion: SPDX-2.2
+DataLicense: CC0-1.0
+DocumentNamespace: http://spdx.org/spdxdocs/spdx-document-xyz
+DocumentName: xyz-0.1.0
+SPDXID: SPDXRef-DOCUMENT
+
+
+# Creation Info
+
+Creator: Organization: Example Inc.
+Creator: Person: Thomas Steenbergen
+Created: 2020-07-23T18:30:22Z
+
+
+# Relationships
+
+Relationship: SPDXRef-Package-xyz CONTAINS SPDXRef-Package-curl
+Relationship: SPDXRef-Package-xyz CONTAINS SPDXRef-Package-openssl
+
+
+# Package
+
+PackageName: xyz
+SPDXID: SPDXRef-Package-xyz
+PackageVersion: 0.1.0
+PackageDownloadLocation: git+ssh://gitlab.example.com:3389/products/xyz.git@b2c358080011af6a366d2512a25a379fbe7b1f78
+FilesAnalyzed: False
+PackageSummary: <text>Awesome product created by Example Inc.</text>
+PackageLicenseDeclared: (Apache-2.0 AND curl AND LicenseRef-Proprietary-ExampleInc)
+PackageLicenseConcluded: NOASSERTION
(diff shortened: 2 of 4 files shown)


spdx/parsers/tagvaluebuilders.py
```diff
@@ -602,6 +602,7 @@ def create_package(self, doc, name):
         name - any string.
         Raise CardinalityError if package already defined.
         """
+        self.reset_package()
```
nicoweidner, line 605: While I am skeptical in general of reusing builder instances (should create a new one for each new instance creation!), this does seem like a reasonable quick fix that does not require larger refactoring.

tests/data/formats/SPDXSBOMExample.tag
```diff
@@ -0,0 +1,72 @@
...
+Creator: Person: Thomas Steenbergen
+Created: 2020-07-23T18:30:22Z
+
+
+# Relationships
+
+Relationship: SPDXRef-Package-xyz CONTAINS SPDXRef-Package-curl
+
+
+# Relationships
```
nicoweidner, line 22: relationships header is duplicated

tests/data/formats/SPDXSBOMExample.tag
```diff
@@ -0,0 +1,72 @@
...
+FilesAnalyzed: False
+PackageFileName: ./libs/openssl
+PackageDescription: <text>OpenSSL is a robust, commercial-grade, full-featured Open Source Toolkit for the Transport Layer Security (TLS) protocol formerly known as the Secure Sockets Layer (SSL) protocol. The protocol implementation is based on a full-strength general purpose cryptographic library, which can also be used stand-alone.</text>
+PackageLicenseDeclared: Apache-2.0
+PackageLicenseConcluded: NOASSERTION
+PackageCopyrightText: <text>copyright 2004-2020 The OpenSSL Project Authors. All Rights Reserved.</text>
+PackageHomePage: https://www.openssl.org/
+
+
+# Extracted Licenses
```
nicoweidner, line 71: This header seems superfluous?

nicoweidner (dismissed):
That should fix the issue, thanks!
I just noticed some tiny details in the new example file that look weird to me

tests/data/formats/SPDXSBOMExample.tag
meretp: I generated this example file with the cli convertor and didn't check the file in detail. So for this example file I will delete the duplicated header and also the superfluous header manually and open a new issue concerning the convertor since this behavior is not expected.

tests/data/formats/SPDXSBOMExample.tag
meretp: done

tests/data/formats/SPDXSBOMExample.tag
meretp: done

nicoweidner: That sounds like a good solution, thanks!

nicoweidner (approved):
Looks good, thanks!

PR #245, opened 2022-10-19
[issue-238] group annotations, reviews and relationships under one headline

### LINKED ISSUE #238: tagvalue writer adds duplicated headlines
This issue came up in #236. Converting the [SPDXSBOMExample.spdx.yml](https://github.com/spdx/tools-python/blob/main/tests/data/formats/SPDXSBOMExample.spdx.yml) file to a .tag file leads to multiple headlines for relationships and a superfluous headline for extracted licenses. This seems to be an issue with the `write_document` method in [tagvalue.py.](https://github.com/spdx/tools-python/blob/main/spdx/writers/tagvalue.py)

meretp:
fixes #238

### DIFF
--- a/spdx/writers/tagvalue.py
+++ b/spdx/writers/tagvalue.py
@@ -23,7 +23,8 @@ class InvalidDocumentError(Exception):
 
     pass
 
-
+def write_separator(out):
+    out.write("\n")
 def write_separators(out):
     out.write("\n" * 2)
 
@@ -64,7 +65,6 @@ def write_review(review, out):
     """
     Write the fields of a single review to out.
     """
-    out.write("# Review\n\n")
     write_value("Reviewer", review.reviewer, out)
     write_value("ReviewDate", review.review_date_iso_format, out)
     if review.has_comment:
@@ -75,7 +75,6 @@ def write_annotation(annotation, out):
     """
     Write the fields of a single annotation to out.
     """
-    out.write("# Annotation\n\n")
     write_value("Annotator", annotation.annotator, out)
     write_value("AnnotationDate", annotation.annotation_date_iso_format, out)
     if annotation.has_comment:
@@ -89,7 +88,6 @@ def write_relationship(relationship_term, out):
     """
     Write the fields of relationships to out.
     """
-    out.write("# Relationships\n\n")
     write_value("Relationship", relationship_term.relationship, out)
     if relationship_term.has_comment:
         write_text_value(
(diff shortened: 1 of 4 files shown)


spdx/writers/tagvalue.py
```diff
@@ -368,7 +374,8 @@ def write_document(document, out, validate=True):
...
 
-    out.write("# Extracted Licenses\n\n")
-    for lic in sorted(document.extracted_licenses):
-        write_extracted_licenses(lic, out)
-        write_separators(out)
+    if document.extracted_licenses:
+        out.write("# Extracted Licenses\n\n")
+        for lic in sorted(document.extracted_licenses):
+            write_extracted_licenses(lic, out)
+            write_separators(out)
```
armintaenzertng, line 381: Why are extracted licenses handled differently than annotations and reviews? (two line breaks vs. one line break between each entry)

spdx/writers/tagvalue.py
```diff
@@ -343,19 +341,27 @@ def write_document(document, out, validate=True):
     write_creation_info(document.creation_info, out)
     write_separators(out)
 
-    # Writesorted reviews
-    for review in sorted(document.reviews):
-        write_review(review, out)
-        write_separators(out)
+    # Write sorted reviews
+    if document.reviews:
+        out.write("# Reviews\n\n")
```
armintaenzertng, line 346: Now that `write_separators` exists, shouldn't we use it here? Or are headlines handled differently?

armintaenzertng (commented):
Thanks for addressing this :)

spdx/writers/tagvalue.py
meretp: We could use it there but in my opinion this would make the code unnecessarily confusing.

spdx/writers/tagvalue.py
meretp: Right, I overlooked that. They should be treated as the others.

spdx/writers/tagvalue.py
armintaenzertng: All right then

spdx/writers/tagvalue.py
```diff
@@ -343,19 +341,27 @@ def write_document(document, out, validate=True):
...
+        for review in sorted(document.reviews):
+            write_review(review, out)
+            write_separator(out)
+        write_separator(out)
```
armintaenzertng, line 350: Reviews and annotations use `write_separator` at the end while all other entries finish with `write_separators`.

Sorry I didn't catch this in the first round!

meretp: This is intended but I also missed this for the extracted licenses. If I would use `write_separators` there, I would get 3 free lines between reviews and the next block which would be annotations. So to be consistent I use `write_separator` here and get 2 free lines between each block.

armintaenzertng: Ah, of course. All right then :)

armintaenzertng (approved):

armintaenzertng (approved):

PR #247, opened 2022-10-20
[issue-246] parse SPDXID for file correctly

### LINKED ISSUE #246: SPDXID from file incorrectly parsed by rdf parser
When parsing a rdf file the SPDXID of a file contains the whole path to the spdxID in the graph. 
For example converting the example file [SPDXRdfExample.rdf](https://github.com/spdx/tools-python/blob/main/tests/data/formats/SPDXRdfExample.rdf) to a .tag file leads to the following output:
```
# File

FileName: Jenna-2.6.3/jena-2.6.3-sources.jar
SPDXID: https://spdx.org/spdxdocs/spdx-example-444504E0-4F89-41D3-9A0C-0305E82C3301#SPDXRef-File1
FileType: ARCHIVE
```

meretp:
fixes #246 

### DIFF
--- a/spdx/parsers/tagvaluebuilders.py
+++ b/spdx/parsers/tagvaluebuilders.py
@@ -1056,6 +1056,7 @@ class FileBuilder(object):
         if self.has_package(doc) and self.has_file(doc):
             if not self.file_spdx_id_set:
                 self.file_spdx_id_set = True
+                spdx_id = spdx_id.split("#")[-1]
                 if validations.validate_file_spdx_id(spdx_id):
                     self.file(doc).spdx_id = spdx_id
                     return True
--- a/spdx/parsers/tagvaluebuilders.py
+++ b/spdx/parsers/tagvaluebuilders.py
@@ -1056,7 +1056,6 @@ class FileBuilder(object):
         if self.has_package(doc) and self.has_file(doc):
             if not self.file_spdx_id_set:
                 self.file_spdx_id_set = True
-                spdx_id = spdx_id.split("#")[-1]
                 if validations.validate_file_spdx_id(spdx_id):
                     self.file(doc).spdx_id = spdx_id
                     return True
--- a/spdx/parsers/rdfbuilders.py
+++ b/spdx/parsers/rdfbuilders.py
@@ -467,6 +467,27 @@ class FileBuilder(tagvaluebuilders.FileBuilder):
         else:
             raise OrderError("File::Notice")
 
+    def set_file_spdx_id(self, doc, spdx_id):
+        """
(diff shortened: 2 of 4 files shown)


spdx/parsers/tagvaluebuilders.py
```diff
@@ -1056,6 +1056,7 @@ def set_file_spdx_id(self, doc, spdx_id):
         if self.has_package(doc) and self.has_file(doc):
             if not self.file_spdx_id_set:
                 self.file_spdx_id_set = True
+                spdx_id = spdx_id.split("#")[-1]
```
zvr, line 1059: If you plan to disregard the part before `#`, you have to at least check that this is exactly the same as the current enclosing identifier.

meretp: Thanks for your review. I don't think that the validation should be part of this PR. In the current implementation this isn't checked for any other ID, e.g. for the [package SPDX ID](https://github.com/spdx/tools-python/blob/c0aa92a3a4779fe189c72b5e1d85a18267f6e45a/spdx/parsers/rdf.py#L338-L341). I would suggest to open another issue for this validation so that the validation is consistent for all SPDXIDs parsed from a rdf-file. Are you okay with that?
Nevertheless I see the point to transfer this rdf-specific logic to the rdf-builder. I will change this.

zvr: Moving it to RDF processing is correct, thanks for that.

I still maintain that the namespace should not be blindly disregarded.

meretp: I opened an issue concerning the validation here #251. It think it makes sense to introduce the valdiation for all values in one PR.  Are you okay with that and would you approve this PR with that?

zvr: I'm fine with handling the namespace elsewhere, especially if it happens for all cases.

But as I wrote in #251, it's not a _validation_. It's a check of whether it can be discarded. In some cases (when it is different than the enclosing namespace), it will have to be retained.

meretp: Yes, I see. I didn't know that there is a difference. But this does also apply to the package SPDX ID, right, e.g. [here](https://github.com/spdx/tools-python/blob/c8ac3630ad5441f860a14ad5115467db99fa5fdc/spdx/parsers/rdf.py#L340)? I will update the open issue.  
With that I think it is best to close this PR without merging and I will change the implementation in #254 so that it is independant from this problem.

zvr: Yes, it might happen in any id you encounter, e.g., document id, snippet id, license id, ...

PR #253, opened 2022-10-26
writers[rdf]: add an option to pass create_doc write

nicoweidner:
Taken over from the original PR by @quaresmajose: https://github.com/spdx/tools-python/pull/199

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,6 +8,7 @@ import unittest
 def test_suite():
     return unittest.TestLoader().discover('tests', pattern='test_*.py')
 
+
 with open('README.md', 'r') as fh:
     long_description = fh.read()
 
@@ -33,6 +34,9 @@ setup(
         'click',
         'pyyaml',
         'xmltodict',
+    ],
+    tests_require=[
+        'pytest',
     ],
     python_requires='>=3.6',
     entry_points={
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -33,6 +33,7 @@ def test_suite():
         'click',
         'pyyaml',
         'xmltodict',
+        'pytest',
```
meretp, line 36: As pytest is only used for testing I am not sure if we should make it a general requirement. For people who only use this tool and don't want to develop anything this package would be installed even if it's never used. For those who want to contribute, the installation is prescribed in the [contributing.md](https://github.com/spdx/tools-python/blob/main/CONTRIBUTING.md#how-to-run-tests).

meretp (commented):
Thank you for adding the unittest. As you already commented this test is a bit clunky but I like the approach to build a minimal document through some helper functions. This is also done [here](https://github.com/spdx/tools-python/blob/c8ac3630ad5441f860a14ad5115467db99fa5fdc/tests/test_document.py#L164-L194) and I think it would be a nice improvement if we somehow merge these kind of functions when the refactoring is done.

setup.py
nicoweidner: Good point, moved it to test dependencies

meretp (approved):

PR #263, opened 2022-11-02
Assorted bug fixes

nicoweidner:
Taken over from https://github.com/spdx/tools-python/pull/207.

There were some conflicts when rebasing on `main`. I tried to make minimal changes, only fixing code that was obviously wrong after resolving conflicts. I will add additional changes in separate commits.

NOTE: Since signoffs of original authors are missing, I will redo this PR from scratch.


spdx/relationship.py
```diff
@@ -112,3 +118,5 @@ def validate_relationship(self, messages):
                 "Relationship type must be one of the constants defined in "
                 "class spdx.relationship.Relationship"
             )
```
meretp, line 120: Why do we use these two functions if the `validate` method only calls the `validate_relationship` method? IMHO we should only use the `validate_relationship` method.

spdx/writers/jsonyamlxml.py
```diff
@@ -475,15 +477,25 @@ def create_document(self):
...
+        document_describes_objects = []
+        for package in unique_doc_packages.values():
             package_info_object = self.create_package_info(package)
             # SPDX 2.2 says to omit if filesAnalyzed = False
             if package.files:
                 package_info_object["files"] = self.create_file_info(package)
-            package_objects.append({"Package": package_info_object})
+            # create a list of the packages' ids for the 'documentDescribes' field
+            package_spdx_id = package_info_object.get("SPDXID")
+            document_describes_objects.append(package_spdx_id)
```
meretp, line 494: With this we will add all packages at document level to the list of documentDescribes SPDXIDs. As far as I understood only the packages which are part of a `DESCRIBES`-Relationship should be listed here. I am not sure if the rebase will tackle this already.

spdx/writers/rdf.py
```diff
@@ -1016,10 +1016,11 @@ def write(self):
...
             self.graph.add(package_triple)
-        """# Add relationship
-        relate_node = self.relationships()
-        relate_triple = (doc_node, self.spdx_namespace.relationship, relate_node)
-        self.graph.add(relate_triple)"""
+        # Add relationship
+        for relate_node in self.relationships():
+            relate_triple = (doc_node, self.spdx_namespace.relationship, relate_node)
+            self.graph.add(relate_triple)
+
```
meretp, line 1023: I did the same as in your comment [here](https://github.com/spdx/tools-python/pull/207#discussion_r964963555) and still got the same result. I think this should be fixed before merging this PR.

meretp (commented):
Thanks for continuing this PR! 
I have a few minor comments, but I'm not sure if the rebase will take care of some of it. I haven't looked at the changes to the example files yet, because I think that will change more with the rebase.

tests/data/doc_write/rdf-simple.json
```diff
@@ -61,8 +61,11 @@
           }, 
           "ns1:fileName": "./some/path/tofile"
         }
-      }, 
+      },
+      "ns1:relationship": {"ns1:Relationship": {"@rdf:about": "SPDXRef-DOCUMENT",
+                                        "ns1:relatedSpdxElement": "SPDXRef-Package",
+                                        "ns1:relationshipType": "DESCRIBES"}},
```
nicoweidner, line 67: A comment about these additions:
This structure looks reasonable to me and is apparently what our rdf writer produces at the moment. It may not be what the spec intends, though:

The structure in e.g. [this example](https://github.com/spdx/spdx-spec/blob/v2.3/examples/SPDXRdfExample-v2.2.spdx.rdf.xml#L1412) is different: The related element is usually expanded (making the whole thing completely unreadable), and there doesn't seem to be a reference to the current element (maybe relationships are put on the originating object instead of the top-level document?).

Considering that there are other oddities in this test file as well (e.g. `referencesFile` and `describesPackage` doesn't exist in example files from the spec or java tools), and the documentation on the desired state is unclear, I'd leave it at this for now. I will ask on the ML what's going on with rdf...

armintaenzertng: As far as I know, elements in rdf files are normally expanded at the first place they appear (which does get quite muddy, yeah).

meretp:
Continued in #270.

PR #273, opened 2022-11-10
add byte range and line ranges to snippet

### LINKED ISSUE #262: snippet class misses range field
According to the spec the [byte range field](https://spdx.github.io/spdx-spec/v2.2.2/snippet-information/#9.3) is required for snippets. The current implementation is missing a field for this in the [snippet class](https://github.com/spdx/tools-python/blob/c8ac3630ad5441f860a14ad5115467db99fa5fdc/spdx/snippet.py#L16-L53). If this field is added, then the parser and writer for all formats must also be adjusted.

meretp:
fixes #262

In this PR the mandatory field `byte range` and the optional field `line range` are added to snippets. Only the jsonyamlxml parser and writer are updated. I didn't add a validation for the mandatory field because the rdf parser/writer and tv parser/writer can't handle these fields yet. That is also why some conversion tests are skipped.


spdx/writers/jsonyamlxml.py
```diff
@@ -370,6 +370,9 @@ def create_snippet_info(self):
             snippet_object["licenseInfoFromSnippet"] = list(
                 map(self.license, snippet.licenses_in_snippet)
             )
+            byte_range = {"endPointer": {"offset": snippet.byte_range[1], "reference": snippet_object["fileId"]},
```
nicoweidner, line 373: Instead of referencing `snippet_object["fileId]"` four times, you could extract `self.spdx_id(snippet.snip_from_file_spdxid)` into a variable in line 369. This would also avoid using a property of the object that we are creating in this method, which I found somewhat confusing when first reading this

spdx/writers/jsonyamlxml.py
```diff
@@ -383,11 +386,18 @@ def create_snippet_info(self):
             if snippet.has_optional_field("license_comment"):
                 snippet_object["licenseComments"] = snippet.license_comment
 
+            if snippet.line_range[0] is not None:
```
nicoweidner, line 389: I missed this in the typing before (sorry!), but I think the range types should be `Optional[Tuple[int, int]]`, and they should be initialized with `None`. It is done the same way for other mandatory and optional fields. Then you should be able to use `has_optional_field` here

tests/test_write_anything.py
```diff
@@ -40,30 +43,40 @@
...
+    in_basename = os.path.basename(in_file_json_yaml_xml)
+    if in_basename == "SPDXSBOMExample.spdx.yml" or in_basename == "SPDXJsonExample2.2.json":
         # conversion of spdx2.2 is not yet done
         return
-    doc, error = parse_anything.parse_file(in_file)
+    write_anything_test(in_basename, in_file_json_yaml_xml, out_format, tmpdir)
+
+
+@pytest.mark.parametrize("out_format_rdf_tag", ['rdf', 'tag'])
+@pytest.mark.parametrize("in_file_rdf_tag", test_files_rdf_tag, ids=lambda x: os.path.basename(x))
```
nicoweidner, line 58: Same here: `out_format` and `in_file` is enough

nicoweidner (commented):
I left some comments, but looks very good overall!

spdx/parsers/jsonyamlxml.py
```diff
@@ -485,6 +487,10 @@ class SnippetParser(BaseParser):
     def __init__(self, builder, logger):
         super(SnippetParser, self).__init__(builder, logger)
 
+    @property
+    def snippet(self) -> Snippet:
+        return self.document.snippet[-1]
```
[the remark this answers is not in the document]
meretp: I totally agree with you, but this was the easisest way for me to not have to use the rdf builder. This is definitely something for the larger refactoring.

tests/test_write_anything.py
meretp: done

tests/test_jsonyamlxml_writer.py
```diff
@@ -1,36 +1,76 @@
...
     assert first_ref.category in written_reference_categories
     assert second_ref.category in written_reference_categories
+
+
+@pytest.mark.parametrize("out_format", ['yaml', 'xml', 'json'])
+def test_snippet_byte_range(temporary_file_path, out_format):
+    snippet = Snippet()
+    snippet.spdx_id = "SPDXRef-Snippet"
+    snippet.byte_range = (310, 420)
+    written_snippet = write_and_parse_snippet(out_format, snippet, temporary_file_path)
```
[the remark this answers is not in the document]
meretp: fixed

tests/test_jsonyamlxml_writer.py
```diff
@@ -1,36 +1,76 @@
...
     written_reference_categories = list(map(lambda x: x.category, written_package.pkg_ext_refs))
     assert first_ref.category in written_reference_categories
     assert second_ref.category in written_reference_categories
+
+
+@pytest.mark.parametrize("out_format", ['yaml', 'xml', 'json'])
+def test_snippet_byte_range(temporary_file_path, out_format):
+    snippet = Snippet()
+    snippet.spdx_id = "SPDXRef-Snippet"
+    snippet.byte_range = (310, 420)
```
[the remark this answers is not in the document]
meretp: done

tests/test_write_anything.py
```diff
@@ -40,30 +43,40 @@
     "SPDXJsonExample2.2.json-tag",
 }
 
+
 @pytest.mark.parametrize("out_format", ['rdf', 'yaml', 'xml', 'json', 'tag'])
-@pytest.mark.parametrize("in_file", test_files, ids=lambda x: os.path.basename(x))
-def test_write_anything(in_file, out_format, tmpdir):
-    in_basename = os.path.basename(in_file)
-    if in_basename == "SPDXSBOMExample.spdx.yml" or in_basename == "SPDXSBOMExample.tag" or in_basename == "SPDXJsonExample2.2.json":
+@pytest.mark.parametrize("in_file_json_yaml_xml", test_files_json_yaml_xml, ids=lambda x: os.path.basename(x))
```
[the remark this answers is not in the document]
meretp: done

spdx/writers/jsonyamlxml.py
meretp: fixed

spdx/writers/jsonyamlxml.py
meretp: I extracted `self.spdx_id(..)` in a variable, not in line 369 but I'm not sure if this line mismatch is due to other changes. Or did you suppose to set this variable in the `__init__` function? This would not work as this variable is different per snippet and `parse_snippets` iterates through all snippets, so the `SnippetParser` isn't initialized per Snippet.

spdx/parsers/jsonyamlxml.py
```diff
@@ -661,6 +668,23 @@ def parse_snippet_license_info_from_snippet(self, license_info_from_snippet):
...
+        - ranges_from_snippet; Python list of dict
+        """
+        if isinstance(ranges_from_snippet, list):
+            for range_dict in ranges_from_snippet:
+                end_pointer = range_dict["endPointer"]
+                start_pointer = range_dict["startPointer"]
+                if "offset" in start_pointer:
+                    self.snippet.byte_range = (int(start_pointer["offset"]), int(end_pointer["offset"]))
+                elif "lineNumber" in start_pointer:
+                    self.snippet.line_range = (int(start_pointer["lineNumber"]), int(end_pointer["lineNumber"]))
```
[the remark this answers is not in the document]
meretp: done

spdx/parsers/jsonyamlxml.py
```diff
@@ -661,6 +668,23 @@ def parse_snippet_license_info_from_snippet(self, license_info_from_snippet):
...
+
+        else:
+            self.value_error("SNIPPET_RANGES", ranges_from_snippet)
```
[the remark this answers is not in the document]
meretp: done

tests/test_document.py
```diff
@@ -390,7 +390,6 @@ def test_write_document_json_multi_package_with_validate(self):
         try:
             temp_dir = tempfile.mkdtemp(prefix='test_spdx')
             result_file = os.path.join(temp_dir, 'spdx-simple-multi-package.json')
-            result_file = 'spdx-simple-multi-package.json'
```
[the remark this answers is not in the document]
meretp: Yeah, I always had this `spdx-simple-multi-package.json` file appearing in my repo and was wondering where this came from and why it wasn't deleted after the tests.

meretp:
tag-value parser and writer are also updated

spdx/snippet.py
```diff
@@ -49,6 +54,8 @@ def __init__(
         self.snip_from_file_spdxid = snip_from_file_spdxid
         self.conc_lics = conc_lics
         self.licenses_in_snippet = []
+        self.byte_range: Tuple[Optional[int], Optional[int]] = byte_range
+        self.line_range: Optional[Tuple[Optional[int], Optional[int]]] = None
```
nicoweidner, line 58: Both of the types should be `Optional[Tuple[int, int]]`, and `byte_range` should be initialized as `None`

tests/test_conversion.py
```diff
@@ -171,12 +174,15 @@ def test_xml_tagvalue(self):
...
 
+
     def test_tagvalue_json(self):
         doc = self.parse_tagvalue_file(utils_test.get_test_loc('formats/SPDXTagExample.tag'))
         filename = get_temp_file('.json')
         self.write_json_file(doc, filename)
         self.parse_json_file(filename)
 
+    @unittest.skip("This fails because we can read/write the mandatory field snippet_byte_range so far only for "
+                   "json-, yaml- or xml-files, not for rdf-/tv-files. https://github.com/spdx/tools-python/issues/274")
```
nicoweidner, line 185: This comment and two more instances further down have to be adapted (they still claim tag/value doesn't work)

tests/test_write_anything.py
```diff
@@ -40,30 +43,37 @@
...
+    if in_basename == "SPDXSBOMExample.spdx.yml" or in_basename == "SPDXJsonExample2.2.json" or in_basename == "SPDXSBOMExample.tag":
         # conversion of spdx2.2 is not yet done
         return
-    doc, error = parse_anything.parse_file(in_file)
+    write_anything_test(in_basename, in_file, out_format, tmpdir)
+
+
+@pytest.mark.parametrize("out_format", ['rdf'])
+@pytest.mark.parametrize("in_file", test_files_rdf, ids=lambda x: os.path.basename(x))
+def test_write_anything_rdf_tag(in_file, out_format, tmpdir):
```
nicoweidner, line 59: `test_write_anything_rdf`

nicoweidner: Also, maybe we should add a comment here explaining why the in/out formats are split like this

tests/test_builder.py
```diff
@@ -806,3 +806,29 @@ def test_snippet_lics_info_order(self):
...
+    def test_snippet_byte_range(self):
+        self.builder.create_snippet(self.document, "SPDXRef-Snippet")
+        self.builder.set_snippet_byte_range(self.document, "310:420")
+
+    @testing_utils.raises(builders.OrderError)
+    def test_snippet_byte_range_order(self):
+        self.builder.set_snippet_byte_range(self.document, "310:420")
+
+    @testing_utils.raises(builders.SPDXValueError)
+    def test_snippet_byte_range_order(self):
```
nicoweidner, line 819: Duplicate test name. This even causes the test not to be executed when you run the whole suite.
Also applies to line range further down

nicoweidner (commented):
Just some minor coments left :slightly_smiling_face:

tests/test_conversion.py
meretp: done

tests/test_write_anything.py
meretp: done

tests/test_builder.py
meretp: fixed

spdx/snippet.py
meretp: done

spdx/parsers/jsonyamlxml.py
```diff
@@ -663,6 +675,61 @@ def parse_snippet_license_info_from_snippet(self, license_info_from_snippet):
         else:
             self.value_error("SNIPPET_LIC_INFO_FIELD", license_info_from_snippet)
 
+    def parse_snippet_ranges(self, ranges_from_snippet: List[Dict]):
```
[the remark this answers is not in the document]
meretp: done

spdx/snippet.py
nicoweidner: Types are still not adjusted (I am not even sure what `Optional[Optional[int], Optional[int]]` means :thinking: )

spdx/writers/jsonyamlxml.py
```diff
@@ -413,6 +414,9 @@ def create_snippet_info(self, annotations_by_spdx_id):
                 snippet_object["licenseInfoFromSnippet"] = list(
                     map(self.license, snippet.licenses_in_snippet)
                 )
+            byte_range = {"endPointer": {"offset": snippet.byte_range[1], "reference": snippet_object["fileId"]},
```
nicoweidner, line 417: Sometimes you're using the `snippet_from_file_spdx_id` local variable, and sometimes reading the same value via `snippet_object["fileId"]`. Both options seem fine to me (slight preference to the former, since we don't need to specify the "fileid" every time), but you shuold be consistent

nicoweidner (approved):

PR #276, opened 2022-11-10
Make 2.3 optional properties optional

### LINKED ISSUE #272: Make properties optional that are not mandatory anymore in 2.3
- [x] package concluded license
  - [x] package declared license
  - [x] package copyright text
  - [x] license information from files
  - [x] package verification code
  - [x] package has files
  - [x] file concluded license
  - [x] license information in file
  - [x] file copyright text
  - [x] snippet concluded license
  - [x] snippet copyright text
  - [x] license information in snippet

armintaenzertng:
fixes #272

only checked and adapted the jsonyamlxml workflow

### DIFF
--- a/spdx/package.py
+++ b/spdx/package.py
@@ -164,7 +164,6 @@ class Package(object):
         self.validate_mandatory_str_fields(messages)
         self.validate_files(messages)
         self.validate_pkg_ext_refs(messages)
-        self.validate_mandatory_fields(messages)
         self.validate_optional_fields(messages)
         messages.pop_context()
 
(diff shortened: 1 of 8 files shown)


spdx/parsers/jsonyamlxml.py
```diff
@@ -658,7 +658,7 @@ def parse_snippet_license_info_from_snippet(self, license_info_from_snippet):
                         self.value_error("SNIPPET_LIC_INFO", lic_in_snippet)
                 else:
                     self.value_error("SNIPPET_LIC_INFO", lic_in_snippet)
-        else:
+        elif license_info_from_snippet is not None:
```
meretp, line 661: If `license_info_from_snippet` is `None` this method is called up unnecessarily. I would suggest to first check in `parse_snippet` if there is a value and if not don't call the method. This applies for all the fields that are mandatory now. I know that this hasn't been done before, so for e.g. packages this would apply as well. I'm not sure if we should change everything in this PR but mybe we could at least start with the parts that are touched here anyway.

meretp (changes requested):
Looks good to me! Only three minor comments, although I am open to discuss whether the parser functions for the mandatory fields should be called (to match the previous style) or not.

spdx/parsers/jsonyamlxml.py
armintaenzertng: I agree that the `None` check is quite hidden there but as I only copied this behaviour from already established cases I propose to leave this refactoring to the big restructure that is coming to the parsers.

spdx/package.py
```diff
@@ -220,20 +203,24 @@ def validate_mandatory_fields(self, messages):
...
     def validate_files(self, messages):
         if self.are_files_analyzed:
-            if not self.files:
-                messages.append(
-                    "Package must have at least one file."
-                )
-            else:
-                for f in self.files:
-                    messages = f.validate(messages)
+            for f in self.files:
```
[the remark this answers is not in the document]
armintaenzertng: done

spdx/parsers/jsonyamlxml.py
nicoweidner: I tend to agree to leave larger refactorings (that should be applied to most of the parser in this case) for later

nicoweidner (approved):
LGTM, thanks!

PR #279, opened 2022-11-10
Recover properties lost during conversion

### LINKED ISSUE #278: Recover missing properties during parsing/writing
When converting from xml to xml, the following properties get lost/changed:
in packages:
- [x] comment

in files:
- [x] attribution texts
- [x] in fileContributors: `file contributor` is converted to `['file contributor']`

armintaenzertng:
fixes #278


meretp (commented):
Looks good to me, I have nothing to comment on.

spdx/parsers/jsonyamlxml.py
```diff
@@ -885,20 +885,27 @@ def parse_file_attribution_text(self, file_attribution_texts):
...
                 self.value_error("FILE_ATTRIBUTION_TEXT", file_attribution_texts)
+        elif isinstance(file_attribution_texts, str):
+            try:
+                return self.builder.set_file_attribution_text(
+                    self.document, file_attribution_texts
+                )
+            except CardinalityError:
+                self.more_than_one_error("FILE_ATTRIBUTION_TEXT")
+            except OrderError:
+                self.order_error("FILE_ATTRIBUTION_TEXT", "FILE_NAME")
```
nicoweidner, line 908: I couldn't find a spot where these errors are raised. If you can't find one either, we can just remove the try/catch blocks

nicoweidner (approved):
Looks like there are some redundant try/catch blocks. I will leave it up to you whether you want to fix those in this PR

spdx/parsers/jsonyamlxml.py
armintaenzertng: I think we will leave it to the larger refactoring

PR #280, opened 2022-11-11
serialize/ deserialize annotations 

### LINKED ISSUE #275: [jsonyamlxml] parse annotations that are nested within packages/snippets/files
For json, yaml and xml-files annotations are subelements of the package/snippet/ file they annotate. Therefore the SPDX identifier reference field is only implicitly given. For now the parser adds the nested annotations at top-level but doesn't parse the SPDXID of the element the annotation originated from.

meretp:
fixes #275

With this PR annotations are attached to the spdx element according to the `annotations.spdx_id` field. Annotations  that refer to external SPDX Elements aren't written as this is currently not allowed in the [json-spec](https://github.com/spdx/spdx-spec/blob/810d229fed8093aae03d9dd42aa0a8174611121a/schemas/spdx-schema.json#L11-L37).

### DIFF
--- a/spdx/writers/jsonyamlxml.py
+++ b/spdx/writers/jsonyamlxml.py
@@ -8,6 +8,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from typing import Dict
 
 from rdflib import Literal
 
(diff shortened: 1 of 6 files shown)


spdx/writers/jsonyamlxml.py
```diff
@@ -122,7 +123,7 @@ def external_reference_as_dict(external_ref: ExternalPackageRef) -> dict:
             external_ref_dict["comment"] = external_ref.comment
         return external_ref_dict
 
-    def create_package_info(self, package):
+    def create_package_info(self, package, annotation_objects):
```
nicoweidner, line 126: I would probably name this `annotations_by_spdx_id`

nicoweidner (dismissed):
Nice, I like your solution!

spdx/writers/jsonyamlxml.py
meretp: renamed

nicoweidner (approved):

PR #284, opened 2022-11-11
Add package properties to tv

### LINKED ISSUE #269: Add missing package properties for SPDX v2.3
See https://github.com/spdx/tools-python/issues/227:
- primary package purpose
- release date
- built date
- valid until date

armintaenzertng:
fixes #269 (again)


spdx/parsers/tagvaluebuilders.py
```diff
@@ -955,6 +960,78 @@ def set_pkg_comment(self, doc, text):
...
+        if not self.package_primary_purpose_set:
+            self.package_primary_purpose_set = True
+            for purpose_enum in PackagePurpose:
+                if purpose == purpose_enum.name:
+                    doc.packages[-1].primary_package_purpose = purpose_enum
+                    return True
+            else:
+                raise SPDXValueError("Package::PrimaryPackagePurpose")
+        else:
+            raise CardinalityError("Package::PrimaryPackagePurpose")
```
nicoweidner, line 979: You can convert this to an early return and reduce indentation:
```
if self.package_primary_purpose_set:
    raise CardinalityError("Package::PrimaryPackagePurpose")
<remaining logic>
```
(in the other methods as well)

spdx/parsers/tagvaluebuilders.py
```diff
@@ -955,6 +960,78 @@ def set_pkg_comment(self, doc, text):
...
+        Raise CardinalityError if more than one purpose is set.
+        Raise SPDXValueError if purpose is unknown.
+        """
+        self.assert_package_exists()
+        if not self.package_primary_purpose_set:
+            self.package_primary_purpose_set = True
+            for purpose_enum in PackagePurpose:
+                if purpose == purpose_enum.name:
+                    doc.packages[-1].primary_package_purpose = purpose_enum
+                    return True
```
nicoweidner, line 975: Not sure if the returned boolean serves any value, but it seems to be the pattern here :man_shrugging:

nicoweidner (changes requested):
Can you add the properties to the files in `tests/data/formats` so it's tested in `test_write_anything`, or will that break something? That seems to be the most comprehensive test we have currently

spdx/parsers/tagvaluebuilders.py
armintaenzertng: there are several analog methods that do not have this `return True`, so I suppose it is unnecessary, but I haven't checked the whole code yet

spdx/parsers/tagvaluebuilders.py
armintaenzertng: that was a lot of work, but it looks much cleaner now! :)

armintaenzertng:
tests are implemented now, too, so this should be ready for another review round :)

spdx/parsers/tagvaluebuilders.py
nicoweidner: Yeah, let's not spend too much time analyzing these builders now

spdx/parsers/tagvaluebuilders.py
```diff
@@ -698,36 +701,35 @@ def set_pkg_down_location(self, doc, location):
...
-                    # validate_pkg_files_analyzed already checked if
-                    # files_analyzed is in ['True', 'true', 'False', 'false']
-                    return True
-                else:
-                    raise SPDXValueError("Package::FilesAnalyzed")
-        else:
+        if self.package_files_analyzed_set:
             raise CardinalityError("Package::FilesAnalyzed")
 
+        if files_analyzed is None or not validations.validate_pkg_files_analyzed(files_analyzed):
```
nicoweidner, line 721: I think the first case is wrong: Before, if `files_analyzed is None`, nothing would happen, right?

spdx/parsers/tagvaluebuilders.py
```diff
@@ -892,18 +895,18 @@ def set_pkg_cr_text(self, doc, text):
...
-                    doc.packages[-1].cr_text = str_from_text(text)
-                else:
-                    doc.packages[-1].cr_text = text  # None or NoAssert
-            else:
-                raise SPDXValueError("Package::CopyrightText")
-        else:
+        if self.package_cr_text_set:
             raise CardinalityError("Package::CopyrightText")
 
+        self.package_cr_text_set = True
```
nicoweidner, line 901: In all other cases, you set the "x_set" boolean _after_ validation. Any reason for doing it before validation here?

spdx/parsers/tagvaluebuilders.py
```diff
@@ -1036,66 +1112,69 @@ def set_file_name(self, doc, name):
...
+            raise SPDXValueError("File::Comment")
+
+        self.file_comment_set = True
+        self.file(doc).comment = str_from_text(text)
+        return True
+
     def set_file_attribution_text(self, doc, text):
         """
         Set the file's attribution text .
+        Raise OrderError if no package or no file defined.
```
nicoweidner, line 1167: Good catch!

spdx/parsers/tagvaluebuilders.py
```diff
@@ -1435,47 +1514,47 @@ def set_snippet_comment(self, doc, comment):
...
             raise CardinalityError("Snippet::SnippetCopyrightText")
 
+        if not validations.validate_snippet_copyright(text, optional=True):
+            raise SPDXValueError("Snippet::SnippetCopyrightText")
+
+        self.snippet_copyright_set = True
+        if isinstance(text, str):
+            doc.snippet[-1].copyright = str_from_text(text)
+        else:
+            doc.snippet[-1].copyright = text  # None or NoAssert
```
nicoweidner, line 1556: nit: No `return True` on this method

nicoweidner (commented):
Wow - when I suggested to use early return, I was referring to the new code you added. Thanks for going over everything and refactoring! That makes the structure a lot clearer, imho, and it also showcases that the logic (and checks) are very similar for all of the properties.

While skimming over the changes, I noticed a few spots that maybe you can have another look at.

spdx/parsers/tagvaluebuilders.py
armintaenzertng: correct, I fixed it with another early `return True` there

spdx/parsers/tagvaluebuilders.py
armintaenzertng: nope, just an oversight

spdx/parsers/tagvaluebuilders.py
armintaenzertng: There might be a few more of these, but this one was glaringly obvious

spdx/parsers/tagvaluebuilders.py
armintaenzertng: Yep, and neither are there in several other of these methods. I didn't bother adding them as their merit seems questionable and everything works fine without them

PR #292, opened 2022-11-15
[issue-291] extract License and its subclasses to a separate file

### LINKED ISSUE #291: extract license class to separate file
To be able to use type hints properly in `file.py` the `License` class should be extracted from `document.py` to a separate file.

meretp:
fixes #291


spdx/file.py
```diff
@@ -43,10 +43,10 @@ class File(object):
     - type: one of FileType.SOURCE, FileType.BINARY, FileType.ARCHIVE
       and FileType.OTHER, optional zero or one.
     - chksum: SHA1, Mandatory one.
-    - conc_lics: Mandatory one. document.License or utils.NoAssert or utils.SPDXNone.
+    - conc_lics: Mandatory one. spdx.license.License or utils.NoAssert or utils.SPDXNone.
```
nicoweidner, line 46: Looks like we sometimes reference it as `spdx.license.License` and sometimes as `license.License`. The second option is shorter (and we already imported `license`), so I'd go with that everywhere

spdx/snippet.py
```diff
@@ -30,10 +30,10 @@ class Snippet(object):
      which this snippet is associated with. Mandatory, one. Type: str.
      - conc_lics: Contains the license the SPDX file creator has concluded as
      governing the snippet or alternative values if the governing license
-     cannot be determined. Mandatory one. Type: document.License or
+     cannot be determined. Mandatory one. Type: License or
```
nicoweidner, line 33: Same as in the other file: Let's try to be consistent and use `license.License` everywhere

spdx/package.py
```diff
@@ -204,31 +204,31 @@ def validate_optional_fields(self, messages):
             )
 
         if self.conc_lics and not isinstance(
-            self.conc_lics, (utils.SPDXNone, utils.NoAssert, document.License)
+            self.conc_lics, (utils.SPDXNone, utils.NoAssert, license.License)
```
nicoweidner, line 207: `license.License` everywhere?

nicoweidner (commented):
I just spotted some inconsistent naming. Thanks for the fix!

spdx/file.py
meretp: You are right, I tried to follow the existing code here. As I understand it, they used `spdx.license.License` when writing an error message, so that the user -probably on top-level- could follow the path. Within the code, so in docstrings it is sufficient to use `license.License` as we already imported it. I don't have a strong opinion about the different usages and the code also isn't consistent here so I will change this for `license.License`. Nevertheless, I think it would be good to make this consistent and decide whether to use `spdx.` everywhere or not.

nicoweidner (approved):

PR #303, opened 2022-11-21
[issue-239] make packages in document optional

### LINKED ISSUE #239: Documents do not have to contain a package anymore
[According to spec](https://spdx.github.io/spdx-spec/v2.3/composition-of-an-SPDX-document/#522-package-information-section), since v2.0 a document does not have to contain a package.

This contradicts the [current validation](https://github.com/spdx/tools-python/blob/a9eeb104e2ccd565af9785d909db5a8777e6dd0a/spdx/document.py#L468).

meretp:
fixes #239

### DIFF
--- a/spdx/document.py
+++ b/spdx/document.py
@@ -175,7 +175,10 @@ class Document(object):
 
     @property
     def files(self):
-        return self.package.files
+        if self.packages:
+            return self.package.files
+        else:
+            return []
 
     @files.setter
     def files(self, value):
@@ -271,17 +274,8 @@ class Document(object):
             messages.append("Document has no creation information.")
 
     def validate_packages(self, messages):
-        if len(self.packages) > 0:
-            for package in self.packages:
-                messages = package.validate(messages)
-        else:
-            messages.append("Document has no packages.")
-
-    def validate_package(self, messages):
-        if self.package is not None:
-            self.package.validate(messages)
-        else:
-            messages.append("Document has no packages.")
+        for package in self.packages:
+            messages = package.validate(messages)
 
     def validate_extracted_licenses(self, messages):
         for lic in self.extracted_licenses:
--- a/spdx/writers/rdf.py
+++ b/spdx/writers/rdf.py
@@ -29,6 +29,7 @@ from spdx.writers.tagvalue import InvalidDocumentError
 
 import warnings
 
+
 class BaseWriter(object):
     """
     Base class for all Writer classes.
@@ -368,7 +369,6 @@ class FileWriter(LicenseWriter):
 
 
 class SnippetWriter(LicenseWriter):
-
     """
     Write spdx.snippet.Snippet
     """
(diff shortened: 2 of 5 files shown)


spdx/writers/rdf.py
```diff
@@ -955,9 +952,8 @@ def pkg_ext_refs(self):
         """
         Return a list of package external references.
         """
-        return map(
-            self.create_package_external_ref_node, self.document.package.pkg_ext_refs
-        )
+        return [self.create_package_external_ref_node(ext_ref) for package in self.document.packages
+                for ext_ref in package.pkg_ext_refs]
```
armintaenzertng, line 956: Normally, external references are part of a package and should not be gathered all in one place, but the rdf writer seems to handle it that way, so we leave that for later.

tests/test_document.py
```diff
@@ -526,9 +535,6 @@ def _get_mini_doc(self,):
         doc.creation_info.add_creator(Tool('ScanCode'))
```
armintaenzertng, line 535: hmm, in `tv-mini.tv` no creator is set... Is this not part of the test?

armintaenzertng (commented):
Thanks, just some short ramblings from my side :)

tests/test_document.py
meretp: Good catch. This is actually not part of the test at the moment. I will delete the creator as it is not needed for writing (better: does not lead to a failure of the test). The creation date must be set because the tv and rdf-writer would fail otherwise and as this is a mandatory field I wouldn't change this.

spdx/writers/rdf.py
meretp: I agree and this is only some kind of workaround to not break the functionality but this will definitely need more work on the rdf writer..

spdx/document.py
```diff
@@ -175,7 +175,10 @@ def package(self, value):
 
     @property
     def files(self):
-        return self.package.files
+        if self.packages:
+            return self.package.files
```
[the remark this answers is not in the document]
meretp: As `self.package` just returns the first element of the list `self.packages` this is a workaround to get the files of the first package. If there are no packages the list is empty and the function would return `None`.  This will be changed once the PR concerning  unpackaged files is merged but for now this is just the workaround to deal with an empty list of packages.

spdx/document.py
```diff
@@ -271,17 +274,8 @@ def validate_creation_info(self, messages):
             messages.append("Document has no creation information.")
 
     def validate_packages(self, messages):
-        if len(self.packages) > 0:
-            for package in self.packages:
+        for package in self.packages:
                 messages = package.validate(messages)
```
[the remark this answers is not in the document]
meretp: fixed

spdx/document.py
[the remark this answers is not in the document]
armintaenzertng: ok, then :)

armintaenzertng (approved):
Thanks for the fixes :)

armintaenzertng (approved):

PR #317, opened 2022-11-25
[issue-261] delete duplicated files

### LINKED ISSUE #261: delete duplicated example files
The example files in  `data/` are the same as in `tests/data/formats`. The latter folder contains more files that are used for testing whereas the first one only contains examples that are used as an entry point to using this tool. I suggest that we delete the files in `data` so that we don't have any duplicate files and point to the files in the test folder as examples in the readme.

meretp:
fixes #261


README.md
```diff
@@ -51,7 +51,7 @@ This is the result of an initial GSoC contribution by @[ah450](https://github.co
 
 1. **PARSER** (for parsing any format):
 * Use `pyspdxtools_parser --file <filename>` where `<filename>` is the location of the file.              
-Try running: `pyspdxtools_parser --file data/SPDXRdfExample.rdf`.
+Try running: `pyspdxtools_parser --file tests/data/formats/SPDXRdfExample.rdf`.
```
nicoweidner, line 54: The example files in `test/data/formats/` are not included in built distributions, only in source distributions (the same is true for the example files in `data/`, for that matter). We should rephrase these comments to something like "If you are using a source distribution, try running..."

nicoweidner (changes requested):
We should rephrase slightly in the readme, since we have to expect users will be using a built distribution, not a source distribution

nicoweidner (approved):

Issue #323, opened 2022-11-28
rdf-parser fails on SPDXRdfExample-v2.2


meretp:
The rdf parser throws an IndexError (`IndexError: list index out of range`), when trying to parse the RdfExample file for spec version 2.2  from the spdx-spec repo ([here](https://github.com/spdx/spdx-spec/blob/development/v2.2.2/examples/SPDXRdfExample-v2.2.spdx.rdf.xml)).

abhi-kr-2100:
I've successfully reproduced this error, and currently working to diagnose and fix it.

abhi-kr-2100:
After some analysis, I've found that the body of the for loop at https://github.com/spdx/tools-python/blob/main/spdx/parsers/rdf.py#L1450 doesn't execute. As a result, `set_ext_doc_id` — the function that populates `doc.ext_document_references` — is not run.

The for loop at https://github.com/spdx/tools-python/blob/main/spdx/parsers/rdf.py#L1459 does run, but since `doc.ext_document_references` is empty, it causes an `IndexError`.

I've fixed this particular problem but the sample file still causes exceptions due to other bugs. I'm making a pull request to get feedback on whether my way of solving these problems is correct since I'm new to SPDX.

armintaenzertng:
The example document `SPDXRdfExample-v2.2.spdx.rdf.xml` can be successfully parsed using the refactored version now found on main.

Issue #295, opened 2022-11-16
rdf support for 2.3


armintaenzertng:
The updated 2.3 data model is only supported with the formats tag-value, xml, json and yaml.
The rdf format support for v2.3 has been postponed.

meretp:
#274 is a subtask of this.

zvr:
Any estimate for this?

I should note that making a "release" (0.7.0) with a title "Support SPDX 2.3" but not actually supporting one of the official formats is not great publicity for the SPDX project.

nicoweidner:
The options were to either release with the rdf caveat, or delay the release for an unknown amount of time as rdf is hard to deal with and the existing code is not in great shape. They were discussed in the weekly public meetings, where it was decided that rdf is less important than json/yaml/xml and less important than tag/value, and supporting rdf can be postponed until major refactoring is done. There is no concrete estimate for when it will be implemented.

armintaenzertng:
Full SPDX 2.3 support is now part of the refactored version found on main.

Issue #227, opened 2022-08-11
Support SPDX 2.3


licquia:
SPDX 2.3 is about to be released; we should support it ASAP.

armintaenzertng:
Fields currently missing in tools-python:

- [ ] files not belonging to a package

in packages:
- [ ] primary package purpose
- [ ] release date
- [ ] built date
- [ ] valid until date

in snippets:
- [ ] byte range (mandatory)
- [ ] line range
- [ ] annotations (probably a parsing issue with (at least) xml)
------------------
Fields that are mandatory in tools-python, but should be optional:
- [ ] packages
- [ ] files in packages
- [ ] package concluded license
- [ ] package declared license
- [ ] package copyright text
- [ ] license information from files
- [ ] package verification code
- [ ] file concluded license
- [ ] license information in file
- [ ] file copyright text
- [ ] snippet concluded license
- [ ] snippet copyright text

These lists are probably not exhaustive. During my experimentation I found lots of issues with the xml parser/writer not detecting fields. These will probably be addressed during the major refactoring.

armintaenzertng:
Issues with the xml parser:

The following fields are now using different tags/values:

in snippets:
- [ ] `fileId` is now `snippetFromFile`
- [ ] `licenseInfoFromSnippet` is now `licenseInfoInSnippets`

in files:
- [ ] file type values are now of the format `SOURCE`, not `fileType_source`
--------------------
During conversion from xml to xml the following fields get lost (whether they are not parsed or not written remains to be determined):

in packages:
- [ ] other checksums than SHA1
- [ ] comment
- [ ] originator

in files:
- [ ] attribution texts
- [ ] other checksums than SHA1
- [ ] in fileContributors: `file contributor` is converted to `['file contributor']`

armintaenzertng:
known shortcomings that we might want to fix later:
- rdf parser, builder and writer
- unpackaged files
- `attributionText` should be a list

armintaenzertng:
The new refactored version now found on main features complete v2.3 support. Please report any problems you may still find with it!

PR #325, opened 2022-11-29
Fix Issue #323

abhi-kr-2100:
The example RDF parser can't parse [the example spec from the `spdx-spec` repo](https://github.com/spdx/spdx-spec/blob/development/v2.2.2/examples/SPDXRdfExample-v2.2.spdx.rdf.xml).

This PR is trying to fix it.

### DIFF
--- a/spdx/parsers/rdfbuilders.py
+++ b/spdx/parsers/rdfbuilders.py
@@ -17,7 +17,7 @@ from spdx import license
 from spdx import package
 from spdx import version
 from spdx.checksum import Checksum, ChecksumAlgorithm
-from spdx.document import Document
+from spdx.document import Document, ExternalDocumentRef
 from spdx.parsers.builderexceptions import CardinalityError
 from spdx.parsers.builderexceptions import OrderError
 from spdx.parsers.builderexceptions import SPDXValueError
@@ -144,6 +144,9 @@ class ExternalDocumentRefBuilder(tagvaluebuilders.ExternalDocumentRefBuilder):
         Set the external document reference's check sum, if not already set.
         chk_sum - The checksum value in the form of a string.
         """
+        if not doc.ext_document_references:
+            doc.add_ext_document_reference(ExternalDocumentRef())
+
         if chk_sum:
             doc.ext_document_references[-1].checksum = Checksum(
                 ChecksumAlgorithm.SHA1, chk_sum
--- a/spdx/parsers/tagvaluebuilders.py
+++ b/spdx/parsers/tagvaluebuilders.py
@@ -181,6 +181,9 @@ class ExternalDocumentRefBuilder(object):
         if not validations.validate_doc_namespace(spdx_doc_uri):
             raise SPDXValueError("Document::ExternalDocumentRef")
 
+        if not doc.ext_document_references:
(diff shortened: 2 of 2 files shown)


abhi-kr-2100:
Please note that even after resolving the exception mentioned in Issue #323, the RDF parser still fails due to other bugs. This PR is hence still in progress.

I opened it to receive feedback on whether the direction I'm taking while trying to debug is acceptable; I'm new to SPDX and may have missed something. Please let me know.

meretp:
Hi @abhi-kr-2100! Thanks for your contribution! 
Unfortunately, I see two problems here:
- Your current solution would possibly overwrite existing information. For example, if one complete `externalDocumentRef` (with ID) has already been parsed and the second `externalDocumentRef` misses an ID (as in the example file) the `checksum` and `spdx_document_uri` fields of the first `externalDocumentRef` would be overwritten with the values from the second `externalDocumentRef`.
-  That said, this issue could be an issue with the provided example file. I opened an [issue](https://github.com/spdx/spdx-spec/issues/816) concering the missing node for the `ExternalDocumentID` in the spdx-spec. The current implementation expects to find an `externalDocumentID`-node (see [here](https://github.com/spdx/tools-python/blob/731e92aec531ad7271ae3f5435dce7b04ec7477a/spdx/parsers/rdf.py#L1450-L1457)) which would then mark the beginning of a new `externalDocumentRef`. As this node is missing in the example file the parser fails when setting the other attributes. Assuming that the example is wrong, you would have to include a check beforehand so that the parser raises an exception if the `externalDocumentID` is not found and terminates and does not raise an IndexError when trying to set the other values. But before you go any further in this, I would recommend to wait for an answer in the open issue in the spec.

nicoweidner:
Hi @abhi-kr-2100 ! Thanks for the contribution!
In addition to what @meretp said, just some general words of advice: We are planning to do the next release very soon, and had already decided to postpone rdf support, since there are too many open issues (see e.g. https://github.com/spdx/tools-python/issues/295 and https://github.com/spdx/tools-python/issues/227#issuecomment-1311715669). After the release, there will be a major break; large parts of the library have to be rewritten to be more modular and maintainable.
What this means: If you find small rdf issues to fix like this one, that's great! But better hold off on any rdf issues that take longer than, say, a few days.

abhi-kr-2100:
@meretp @nicoweidner Thank you for the advice and pointers. As a beginner to open source contributions, I really appreciate it. I'll be looking around and see what I can meaningfully contribute to. :)

nicoweidner:
@abhi-kr-2100 If you want to get involved in where development is going currently, you're welcome to join our weekly call.  It's tomorrow, Thursday, 5pm - 5.30pm GMT, https://zoom.us/j/98741582779. The regular time is half an hour earlier, but the first Thursday every month is shifted because of a conflict. It's open to everyone, and meant as an informal way to sync on current topics and where development should focus next.

I hope the current transitional phase will be over soon. Once we have the new data model merged (currently on branch https://github.com/spdx/tools-python/tree/refactor-python-tools), there will be plenty of areas to get involved in - for example, porting the writers over to the new data model.

abhi-kr-2100:
@nicoweidner That sounds awesome. I'm looking forward to it. Thanks once again!

PR #327, opened 2022-11-29
[issue-326] fix output for licenses that are SPDXNone or NoAssert-objects

### LINKED ISSUE #326: fix cli parser output for SPDXNone und NoAssert
The objects in the `licenses_in_file` field of  `file` objects can either be `License` objects, `SPDXNone` or `NoAssert` objects (see the [validation](https://github.com/spdx/tools-python/blob/731e92aec531ad7271ae3f5435dce7b04ec7477a/spdx/file.py#L184-193)). When writing this info to the console ([here](https://github.com/spdx/tools-python/blob/731e92aec531ad7271ae3f5435dce7b04ec7477a/spdx/cli_tools/parser.py#L89-90)) the cli parser tries to access the field `identifier` which is only valid for a `License` object. The parser needs to be fixed here. 
Additionally one should check, if there are more fields (in addition to `file.licenses_in_file`) where this applies.

meretp:
fixes #326 

### DIFF
--- a/spdx/cli_tools/parser.py
+++ b/spdx/cli_tools/parser.py
@@ -12,6 +12,8 @@
 # limitations under the License.
 
 import os
+
+from spdx import utils
 from spdx.parsers.parse_anything import parse_file
 import spdx.file as spdxfile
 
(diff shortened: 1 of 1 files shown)


meretp:
> Did you check for other occurrences as you stated in the issue?
Yes, but I couldn't find any.

Issue #324, opened 2022-11-28
Add SPDX 2.2 and 2.3 examples to test files


nicoweidner:
SPDX 2.2 and 2.3 should now be fully supported apart from rdf. Examples from the spdx-spec repo should be added to https://github.com/spdx/tools-python/tree/main/tests/data/formats, so that they are picked up by https://github.com/spdx/tools-python/blob/main/tests/test_write_anything.py. The conversions marked as unstable in that test should be checked and enabled if possible.

PR #332, opened 2022-11-29
[issue-324] Add examples from spdx-spec, various bugfixes

nicoweidner:
Fixes https://github.com/spdx/tools-python/issues/324


tests/utils_test.py
```diff
@@ -177,7 +177,7 @@ def load_and_clean_json(location):
     data = json.loads(content)
 
     if 'creationInfo' in data:
-        del(data['creationInfo'])
+        del (data['creationInfo'])
```
armintaenzertng, line 180: Is `del` special somehow to earn a space between the next bracket?

tests/utils_test.py
```diff
@@ -350,20 +355,20 @@ def package_to_dict(cls, package):
...
             ('sourceInfo', package.source_info),
-            ('downloadLocation', package.download_location),
+            ('downloadLocation', str(package.download_location)),
             ('homepage', package.homepage),
             ('originator', cls.entity_to_dict(package.originator)),
             ('supplier', cls.entity_to_dict(package.supplier)),
             ('licenseConcluded', cls.license_to_dict(package.conc_lics)),
             ('licenseDeclared', cls.license_to_dict(package.license_declared)),
-            ('copyrightText', package.cr_text),
+            ('copyrightText', str(package.cr_text)),
```
armintaenzertng, line 364: Are these additional `str` used to handle `NONE` or `NO_ASSERTION`? If so, aren't there other fields that need the same treatment, like `homepage`?

armintaenzertng (commented):
Looks good, just a few remarks :)

tests/utils_test.py
nicoweidner: Yeah, looks like it's a special reserved keyword. I didn't even notice this, autoformat did it...

tests/utils_test.py
nicoweidner: Maybe. These are the ones that I saw in some cases. I have to admit, I didn't fully understand the issue here. It showed some differences where `NOASSERTION` was inside quotation marks on one side, and without quotation marks on the other. I debugged and noticed that in the latter case, e.g. `package.cr_text` was a `NoAssert` instance. However, its `str` and `repr` methods do return a string, so I am not quite sure why casting it explicitly as string made a slight difference.

That said, all of this is test utility code that will probably be replaced, so I was fine with just fixing the issues that I observed in any halfway-reasonable way.

armintaenzertng: fair enough

armintaenzertng (approved):

armintaenzertng (approved):

Issue #338, opened 2022-12-01
Update changelog for 0.7.0 release


nicoweidner:
A changelog was started a while ago and should be updated for the 0.7.0 release: https://github.com/spdx/tools-python/blob/main/CHANGELOG.rst. We can also use it for the release notes.

PR #339, opened 2022-12-01
[issue-338] Update changelog for 0.7.0

nicoweidner:
In principle, I like to include a `Bugfixes` section on the same level as the `New features` section. But since this release spans such a long time, I decided to stick to a higher-level description and not create a bugfixes section.

Fixes https://github.com/spdx/tools-python/issues/338


CHANGELOG.md
```diff
@@ -0,0 +1,63 @@
...
+* Extended the contribution guidelines
+* Improved tag/value output structure
+* Added .editorconfig and pyproject.toml
+* Improved handling of JSON-specific properties `documentDescribes` and `hasFiles`
+* Fixed annotation handling for the JSON and Tag/Value formats
+* Free form text values in Tag/Value no longer require `<text>` tags if they don't span multiple lines
+
+### Contributors
+
+Thank you to these contributors:
```
armintaenzertng, line 35: This sounds weird to me. Maybe something like "This release was made possible by the following contributors, thank you very much!" or something...?

CHANGELOG.md
```diff
@@ -0,0 +1,63 @@
...
+
+
+## v0.7.0 (TBD)
+
+Starting a Changelog.
+
+### New features and changes
+
+* Dropped Python 2 support. Python >= 3.6 is now required.
+* Added new LicenseListVersion tag.
```
armintaenzertng, line 11: This seems comparatively minor to stand at position 2 in the list. Also, it's probably only a part of the "updated the tools"-point

CHANGELOG.md
```diff
@@ -0,0 +1,63 @@
...
+* Added `pyspdxtools_convertor` and `pyspdxtools_parser` CLI scripts. See [the readme](README.md) for usage instructions.
```
armintaenzertng, line 12: Didn't we just rename them instead of adding them?

Also, we improved the output of the CLI parser (not saying it is perfect now, though)

CHANGELOG.md
```diff
@@ -0,0 +1,63 @@
...
+* Updated the tools to support SPDX versions up to 2.3 and to conform with the specification. Apart from many bugfixes 
+  and new properties, some of the more significant changes include:
+    * Support for multiple packages per document
+    * Support for multiple checksums for packages and files
+    * Support for files outside a package
+    * Removed example documents from production code. Added additional up-to-date examples to test files.
```
armintaenzertng, line 18: This sounds like it's own point, not part of the sub list

armintaenzertng (commented):
Do we want to make the rdf shortcoming transparent in the changelog?

CHANGELOG.md
nicoweidner: I don't think being able to specify the license list version is a 2.2 or 2.3 feature, or is it? But I'll move it further down

CHANGELOG.md
nicoweidner: No, they were not included at all in the previous release

CHANGELOG.md
nicoweidner: Yeah, fair enough. The reason it was there is that the new files are 2.2/2.3-specific. But it makes more sense as a point of its own

armintaenzertng (approved):
Thanks :)

PR #340, opened 2022-12-01
Refactor the data model to use dataclass, properties and typeguard

### LINKED ISSUE #321: Add setters with type checks in new data model
We should add setters to the new data model that include a type check, to make sure that instances are "type-valid". While this is already useful in itself, it also helps avoid doing type checks in multiple places (parser and validation).

armintaenzertng:
fixes #321


armintaenzertng:
wrong target branch

PR #351, opened 2022-12-07
[fix] change annotator type to actor in constructor

armintaenzertng:


meretp (approved):
Good catch!

PR #352, opened 2022-12-07
[issue-243] delete duplicated license list version field

meretp:
I decided to keep the input argument for `license_list_version` for the `document` object for backwards compatibility and redirect the input to the `creation_info` field. The current parsers and writers only use the `license_list_version` field from the `creation_info` so nothing needs to be changed here.

### DIFF
--- a/spdx/document.py
+++ b/spdx/document.py
@@ -120,7 +120,6 @@ class Document(object):
         self.version = version
         self.data_license = data_license
         self.name = name
-        self.license_list_version=license_list_version
         self.spdx_id = spdx_id
         self.ext_document_references = []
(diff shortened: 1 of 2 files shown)


spdx/document.py
```diff
@@ -120,7 +120,6 @@ def __init__(
         self.version = version
         self.data_license = data_license
         self.name = name
-        self.license_list_version=license_list_version
```
armintaenzertng, line 123: If we want backwards compatibility, we shouldn't erase fields from classes that people might be using.

armintaenzertng (approved):
Nice change that hopefully doesn't break anything ;)

PR #361, opened 2022-12-08
[issue-358] remove legacy structure

nicoweidner:
Note: This is a pretty radical change. The idea behind it is that we can get the `refactor-python-tools` branch merged to `main` now that the 0.7.0 release is out, and then gradually add the various parsers and writers. The previous state can always be fetched from the git history (which should prove helpful for the writers in particular)


armintaenzertng:
We should also remove the old tests lest they all fail.

src/clitools/parser.py
```diff
@@ -0,0 +1,123 @@
...
+
+    # Parse document and set as doc here
+    # First one to implement is the Json parser: https://github.com/spdx/tools-python/issues/305
+
+    print("doc comment: {0}".format(doc.comment))
+    print("Creators:")
+    for c in doc.creation_info.creators:
+        print("\t{0}".format(c))
+    print("Document review information:")
+    for review in doc.reviews:
```
meretp, line 40: As we are deleting the old data model, I think we should also delete these old lines of code. They refer to fields that no longer exist. We could keep some kind of "rough skeleton" (e.g. print `creation_info`, `packages` and `files` or whatever) to keep the "main idea" of this script but I would delete most of the code.

meretp (commented):
Nice breaking change! Just a small remark from my side.

src/clitools/parser.py
nicoweidner: Good catch! I didn't look at these too closely (it's mostly just copied over), but we can remove it right away since it doesn't make sense in the new data model

meretp (approved):

PR #365, opened 2022-12-09
add class name to error messages from setter/ getter

meretp:


armintaenzertng (approved):
Nice improvement also with the indication of setting or getting the field :)

PR #370, opened 2022-12-12
[fix] remove leftover legacy files. Add `NOASSERTION` as possible value…

armintaenzertng:
… for extracted licensing info license name


meretp (approved):
Good catch, I think it is good that we can do without the additional license!

PR #401, opened 2023-01-02
[issue-388] Update CLI tools for the new parsers/writers

### LINKED ISSUE #388: add cli parser for json files
As #366 is merged, we can add a cli parser for json-files. subtask of #383

armintaenzertng:
fixes #388 

parsing output uses the tag-value writer to output to stdout


src/clitools/parser.py
```diff
@@ -11,29 +11,49 @@
...
+        return
+
+    if printout:
+        write_document(document, sys.stdout)
+        print("")
+
+    if validate:
+        validation_messages: List[ValidationMessage] = validate_full_spdx_document(document, version)
+        if validation_messages:
+            print("The document is invalid. The following issues have been found:")
```
maxhbr, line 52: If not valid, the cli should `exit` with `1`

src/clitools/parser.py
```diff
@@ -11,29 +11,49 @@
...
-    # Print all document properties - or possibly a selection of them. Should be human-readable, so using indentation
-    # for nested properties is probably a good idea.
+    try:
+        document: Document = parse_file(file)
+    except NotImplementedError as err:
+        print(err.args[0])
+        print("Please note that this project is currently undergoing a major refactoring and therefore missing "
+              "a few features which will be added in time.\n"
+              "In the meanwhile, please use the current PyPI release version 0.7.0.")
+        return
```
maxhbr, line 43: in case of error, it should also `exit` with `1`

src/clitools/parser.py
```diff
@@ -11,29 +11,49 @@
...
-    # Parse document
-    # First one to implement is the Json parser: https://github.com/spdx/tools-python/issues/305
-
-    # Print all document properties - or possibly a selection of them. Should be human-readable, so using indentation
-    # for nested properties is probably a good idea.
+    try:
+        document: Document = parse_file(file)
+    except NotImplementedError as err:
+        print(err.args[0])
+        print("Please note that this project is currently undergoing a major refactoring and therefore missing "
```
maxhbr, line 40: maybe link to the issues?

src/clitools/convertor.py
```diff
@@ -11,100 +11,33 @@
...
-        infile, outfile = determine_infile_and_outfile(infile, outfile, src, from_, to)
-    except ValueError as err:
-        print(err)
-        print_help_msg(main)
+        document: Document = parse_file(infile)
+
+        write_file(document, outfile)
+    except NotImplementedError as err:
+        print(err.args[0])
+        print("Please note that this project is currently undergoing a major refactoring and therefore missing "
```
maxhbr, line 36: also here: on error exit with code 1

src/clitools/parser.py
```diff
@@ -11,29 +11,49 @@
...
+from src.writer.tagvalue.tagvalue_writer import write_document
 
 
 @click.command()
 @click.option("--file", prompt="File name", help="The file to be parsed")
-@click.option("--force", is_flag=True, help="print information even if there are some parsing errors")
-def main(file, force):
+@click.option("--version", prompt="SPDX version", help="The SPDX version to be used during validation")
+@click.option("--validate", is_flag=True, help="validate the provided document")
+@click.option("--printout", is_flag=True, help="print the parsed document to stdout in tag-value format")
```
maxhbr, line 30: this is a pure debug option?

An often seen convention, to print to stdout, is to use the argument `-`. So `-o -` implies output is stdout.

src/clitools/parser.py
```diff
@@ -11,29 +11,49 @@
...
+def main(file, version, validate, printout):
```
maxhbr, line 31: why is there a parser and a converter? I think they both could be the same tool.

src/formats.py
```diff
@@ -0,0 +1,36 @@
...
+
+from src.parser.error import SPDXParsingError
+
+
+class FileFormat(Enum):
+    JSON = auto()
+    YAML = auto()
+    XML = auto()
+    TAG_VALUE = auto()
+    RDF = auto()
```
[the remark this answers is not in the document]
armintaenzertng: We use these interchangeably, but `RDF_XML` is more precise, I guess. Fixed.

src/clitools/parser.py
armintaenzertng: Parser and converter have been merged into a single tool.

src/clitools/parser.py
armintaenzertng: Parser and converter have been merged into a single tool.

src/clitools/parser.py
armintaenzertng: done

src/clitools/convertor.py
armintaenzertng: done

src/clitools/parser.py
armintaenzertng: done

Issue #320, opened 2022-11-25
Use version class for spdx version in the new data model


armintaenzertng:
Also add a third number for patch version.

Then we can compare versions with each other, which is especially helpful with version-specific validations.

armintaenzertng:
We should consider using the `semantic_version` package for this. Especially since it will probably be used in the SPDX3 model, too.

armintaenzertng:
As we only support `SPDX-2.2` and `SPDX-2.3`, I decided that a specific version class would be too much trouble for what it's worth.

PR #404, opened 2023-01-03
[issue-403] change spdx version handling during validation

### LINKED ISSUE #403: Use document SPDX version as default for validation
An optional `version` parameter can be given to the document validator to overwrite the document version. In case of a discrepancy, a validation error will be logged.

armintaenzertng:
fixes #403 

### DIFF
--- a/src/validation/creation_info_validator.py
+++ b/src/validation/creation_info_validator.py
@@ -19,23 +19,10 @@ from src.validation.uri_validators import validate_uri
 from src.validation.validation_message import ValidationMessage, ValidationContext, SpdxElementType
 
 
-def validate_creation_info(creation_info: CreationInfo, spdx_version: str) -> List[ValidationMessage]:
+def validate_creation_info(creation_info: CreationInfo) -> List[ValidationMessage]:
     validation_messages: List[ValidationMessage] = []
 
     context = ValidationContext(spdx_id=creation_info.spdx_id, element_type=SpdxElementType.DOCUMENT)
-
-    if not re.match(r"^SPDX-\d+.\d+$", creation_info.spdx_version):
-        validation_messages.append(
-            ValidationMessage(
-                f'spdx_version must be of the form "SPDX-[major].[minor]" but is: {creation_info.spdx_version}',
-                context
-            )
-        )
-    elif spdx_version != creation_info.spdx_version:
-        validation_messages.append(
-            ValidationMessage(f"provided SPDX version {spdx_version} does not match "
-                              f"the document's SPDX version {creation_info.spdx_version}", context)
-        )
 
     if creation_info.spdx_id != "SPDXRef-DOCUMENT":
         validation_messages.append(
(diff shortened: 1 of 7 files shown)


src/validation/document_validator.py
```diff
@@ -24,10 +24,13 @@
 from src.validation.validation_message import ValidationMessage, ValidationContext, SpdxElementType
 
 
-def validate_full_spdx_document(document: Document, spdx_version: str) -> List[ValidationMessage]:
+def validate_full_spdx_document(document: Document, spdx_version: str = None) -> List[ValidationMessage]:
     validation_messages: List[ValidationMessage] = []
 
-    validation_messages.extend(validate_creation_info(document.creation_info))
+    if not spdx_version:
+        spdx_version = document.creation_info.spdx_version
```
maxhbr, line 31: instead of setting the version here, I think it would be cleaner to make the version optional in the `validation_messages.extend(validate_creation_info(` call and only assert on it, if it is set.

src/validation/creation_info_validator.py
```diff
@@ -19,7 +19,7 @@
 from src.validation.validation_message import ValidationMessage, ValidationContext, SpdxElementType
 
 
-def validate_creation_info(creation_info: CreationInfo) -> List[ValidationMessage]:
+def validate_creation_info(creation_info: CreationInfo, spdx_version: str) -> List[ValidationMessage]:
```
maxhbr, line 22: as mentioned before, I think the version should be optional here

src/validation/relationship_validator.py
```diff
@@ -43,9 +43,9 @@ def validate_relationship(relationship: Relationship, document: Document, spdx_v
         for message in messages:
             validation_messages.append(ValidationMessage(message, context))
 
-    if spdx_version != "2.3":
+    if spdx_version != "SPDX-2.3":
```
maxhbr, line 46: might be good to put `"SPDX-2.3"` into an ENUM

armintaenzertng: Actually, my plan was to convert `spdx_version` to a proper version field to then be able to properly implement conditions like "for all version below 2.3". See https://github.com/spdx/tools-python/issues/320

src/validation/document_validator.py
armintaenzertng: I agree that it should be optional in `validate_creation_info()`, but we have to set the version here, nevertheless, as it is a mandatory field for other validations like `validate_relationships()` (and more to come later).

maxhbr: you could still move the spdx version validation up, to have it cleaner

maxhbr: Its just a minor improvement, but:
```suggestion
    validation_messages.extend(validate_creation_info(document.creation_info, spdx_version))
    
    if not spdx_version:
        spdx_version = document.creation_info.spdx_version
```

armintaenzertng: I have extracted the version validation to be handled before everything else, now.

PR #410, opened 2023-01-04
[issue-267] Add GitHub Actions workflow

### LINKED ISSUE #267: Move CI to Github Actions
Currently, Appveyor is broken, and supposedly it is not the first time that it's unstable.
Many people consider Github Actions a better CI solution than CircleCI.

For these two reasons, I'd suggest to move all CI checks to Github Actions. It seems to support runners for Linux, MacOS and Windows: https://github.com/actions/runner-images

armintaenzertng:
fixes #267

removing the CircleCI workflow

this makes even Windows working again :)


armintaenzertng:
They do run on my branch: https://github.com/armintaenzertng/tools-python/actions

I have added `pull_request` as a trigger for the workflow, but I think we have to merge this first, as currently the pipeline is still trying to use CircleCI.

PR #418, opened 2023-01-10
add dummy config.yml to pass pipeline

meretp:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,15 @@
+# Empty Circle CI configuration file to make pipeline pass
+
+version: 2.1
+
+jobs:
+  empty-job:
+    docker:
+      - image: python:3.11
+    steps:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,15 @@
+# Dummy Circle CI configuration file to make pipeline pass
+
+version: 2.1
+
+jobs:
+  dummy-job:
+    docker:
+      - image: python:3.11
+    steps:
+      - run: echo "This is a dummy job"
```
maxhbr, line 10:
```suggestion
      - run: echo "Empty Job to make CircleCI green, we switched to https://github.com/spdx/tools-python/actions"
```

maxhbr (changes requested):
dummy is not the best name

maxhbr (approved):
LGTM

PR #419, opened 2023-01-10
also use the src-layout for tests

meretp:
in addition to #416 


armintaenzertng (approved):
Thanks :)
This will already be the first rebase of the 3.0 branch :D

PR #421, opened 2023-01-11
[issue-420] fix order of mocks to match input arguments

### LINKED ISSUE #420: fix order of mocks in tests for the refactored data model
As described [here](https://docs.python.org/3/library/unittest.mock.html#quick-guide) mocks are passed in to the decorated function  in the order they applied, so bottom up. Therefore the order in the current tests need to be changed (e.g. [here](https://github.com/spdx/tools-python/blob/11727e27f1190e81a1fb6dbb5f3c46b87701dfd7/tests/model/test_package.py#L13-L16). This doesn't lead to faliing tests since `typeguard` ignores instances of `Mock` or `MagicMock` (see [here](https://typeguard.readthedocs.io/en/stable/userguide.html#support-for-mock-objects)).

meretp:
fixes #420

### DIFF
--- a/tests/model/test_creation_info.py
+++ b/tests/model/test_creation_info.py
@@ -7,8 +7,8 @@ from spdx.model.document import CreationInfo
 from spdx.model.version import Version
 
 
-@mock.patch('spdx.model.actor.Actor', autospec=True)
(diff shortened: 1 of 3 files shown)


maxhbr (approved):
LGTM

PR #424, opened 2023-01-14
Reimplemented parse_anything.py to handle SBOMs with incorrect file extensions. 

linynjosh:
I tried parsing the SBOMs in the SPDX DocFest 20220127 and more than two thirds of the documents could not be parsed due to their file extensions. This was because most files had the extension .spdx and the parse_anything function assumed that every .spdx file should be parsed as a .tag file. However, there were many .spdx files that were not tag files and an error was thrown before they were parsed. Furthermore, there were examples of files with incorrect file extensions which prevented them from being parsed at all. I discussed with @licquia during the summer and he suggested that I try all the parsers for each file until the correct parser matches the type of the file. Below is piece of code where most .spdx documents fail on.
Screen Shot 2023-01-14 at 4 09 53 AM

### DIFF
--- a/spdx/parsers/parse_anything.py
+++ b/spdx/parsers/parse_anything.py
@@ -22,33 +22,25 @@ from spdx.parsers.builderexceptions import FileTypeError
 
 
 def parse_file(fn):
-    buildermodule = jsonyamlxmlbuilders
-    read_data = False
-    if fn.endswith(".rdf") or fn.endswith(".rdf.xml"):
-        parsing_module = rdf
-        buildermodule = rdfbuilders
-    elif fn.endswith(".spdx"):
-        parsing_module = rdf
-        buildermodule = rdfbuilders
-    elif fn.endswith(".tag"):
-        parsing_module = tagvalue
-        buildermodule = tagvaluebuilders
-        read_data = True
-    elif fn.endswith(".json"):
-        parsing_module = jsonparser
-    elif fn.endswith(".xml"):
-        parsing_module = xmlparser
-    elif fn.endswith(".yaml") or fn.endswith(".yml"):
-        parsing_module = yamlparser
-    else:
-        raise FileTypeError("FileType Not Supported" + str(fn))
+    buildermodules = [rdfbuilders, jsonyamlxmlbuilders, jsonyamlxmlbuilders, jsonyamlxmlbuilders, tagvaluebuilders]
+    parsing_modules = [rdf, xmlparser, yamlparser, jsonparser, tagvalue]
+    read_datas = [False, False, False, False, True]
+    for i in range(len(buildermodules)):
+        parsing_module = parsing_modules[i]
+        buildermodule = buildermodules[i]
+        read_data = read_datas[i]
+        try:
+            p = parsing_module.Parser(buildermodule.Builder(), StandardLogger())
+            if hasattr(p, "build"):
+                p.build()
+            with open(fn) as f:
+                if read_data:
+                    data = f.read()
+                    return p.parse(data)
+                else:
+                    return p.parse(f)
+        except:
+            if i == len(buildermodules) - 1:
+                raise FileTypeError("FileType Not Supported" + str(fn))
 
-    p = parsing_module.Parser(buildermodule.Builder(), StandardLogger())
-    if hasattr(p, "build"):
-        p.build()
-    with open(fn) as f:
-        if read_data:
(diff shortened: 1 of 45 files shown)


spdx/parsers/parse_anything.py
```diff
@@ -22,33 +22,25 @@
...
+            with open(fn) as f:
+                if read_data:
+                    data = f.read()
+                    return p.parse(data)
+                else:
+                    return p.parse(f)
+        except:
+            if i == len(buildermodules) - 1:
+                raise FileTypeError("FileType Not Supported" + str(fn))
+
```
goneall, line 45: This code will work, but I would personally find it more readable if on the exception you just continued and raise the error after the for loop since if you made it there, all file types would fail parsing.

I would also change the message to something like "The file cannot be parsed by the RDF, XML, YAML, JSON or Tag/Value parsers".

spdx/parsers/parse_anything.py
```diff
@@ -22,33 +22,25 @@
...
-        parsing_module = jsonparser
-    elif fn.endswith(".xml"):
-        parsing_module = xmlparser
-    elif fn.endswith(".yaml") or fn.endswith(".yml"):
-        parsing_module = yamlparser
-    else:
-        raise FileTypeError("FileType Not Supported" + str(fn))
+    buildermodules = [rdfbuilders, jsonyamlxmlbuilders, jsonyamlxmlbuilders, jsonyamlxmlbuilders, tagvaluebuilders]
+    parsing_modules = [rdf, xmlparser, yamlparser, jsonparser, tagvalue]
+    read_datas = [False, False, False, False, True]
```
goneall, line 27: Not a necessary change from my perspective, but it would be nice to attempt parsing based on the file type as the first attempt just for performance reasons.

You could do something like leaving in the code for setting the parsing module and attempting that parse first and only looping through if the parsing_module is not set or if the parse fails.

goneall (changes requested):
Thanks @linynjosh for the code updates!  This should help especially when using it to analyze a larger dataset.

Just a couple of suggestions - neither are required from my perspective.

I also noticed we have a merge conflict, so you may need to merge your changes into the latest code.

spdx/parsers/parse_anything.py
```diff
@@ -22,33 +22,25 @@
...
+            p = parsing_module.Parser(buildermodule.Builder(), StandardLogger())
+            if hasattr(p, "build"):
+                p.build()
+            with open(fn) as f:
+                if read_data:
+                    data = f.read()
+                    return p.parse(data)
+                else:
+                    return p.parse(f)
+        except:
```
meretp, line 42: This would basically `except` everything, I am not sure if it is possible, but I would try to make sure that this only excepts "basic" failure if the parser doesn't match the content of the file. So e.g. if we try to use the `rdf` parser for a file which has a `.json` ending but is in `tag-value` format. This should not except any other `Errrors` that may occur during parsing, otherwise these errors would simply no longer be visible.

meretp (changes requested):
Thanks for your PR @linynjosh! 
After some reflection and with regard to @goneall's comment, I see the benefit of this new feature and support this in general. (I will create an issue so that we also implement this feature in the refactored data model). 

However, I think we should change the test-setup: Instead of adding various new files to the repo (which in the end are only duplications of the existing ones and don't add any value), I would suggest to implement a fixture that only temporarily changes the ending of the file under test. With that we could use the existing test files in format (see also the third bullet point in this [comment](https://github.com/spdx/tools-python/pull/223#discussion_r963525684)).
Additionally, we could make one "positive" test that passes if the files could be parsed with the correct ending, one test that passes if files with wrong ending can be parsed and potentially also one test that passes if the `FileTypeError` is correctly raised. Overall, we should ensure that `FileTypeError` is only raised if the content of the file doesn't match any of JSON/ YAML/ XML/ RDF or TV as commented below.

Edit: One additional remark on the merge conflict and the failing pipeline: This definitely needs to be fixed before merging this PR but at least the merge conflicts shouldn't be complicated. As I can't see the Details for the failing pipeline I am not sure about this, but this will probably also be fixed with the requested changes.

armintaenzertng:
The main branch has been replaced with the refactored version of the tools-python. The version this PR refers to can be found in the [version/v0.7 branch](https://github.com/spdx/tools-python/tree/version/v0.7). Please reopen this PR against that branch while addressing the requested changes from here.

PR #430, opened 2023-01-18
move typing to common module to prepare for the prototype of spdx 3.0

meretp:


src/spdx/parser/jsonlikedict/dict_parsing_functions.py
```diff
@@ -12,7 +12,7 @@
 
 from spdx.model.spdx_no_assertion import SpdxNoAssertion
 from spdx.model.spdx_none import SpdxNone
-from spdx.model.typing.constructor_type_errors import ConstructorTypeErrors
+from common.typing.type_checks import ConstructorTypeErrors
```
armintaenzertng, line 15: Why are `ConstructorTypeErrors` imported from `type_checks` now? Does this even work?

src/spdx/parser/jsonlikedict/relationship_parser.py
```diff
@@ -11,7 +11,7 @@
 from typing import Dict, List, Optional
 
 from spdx.model.relationship import Relationship, RelationshipType
-from spdx.model.typing.constructor_type_errors import ConstructorTypeErrors
+from common.typing.type_checks import ConstructorTypeErrors
```
armintaenzertng, line 14: Same as above

src/spdx/parser/jsonlikedict/dict_parsing_functions.py
meretp: Good catch! Fixed this.

armintaenzertng (approved):
Looks good to me now :)

PR #433, opened 2023-01-19
[issue-431] make Element, Artifact, SpdxCollection and IntegrityMetho…

### LINKED ISSUE #431: use abstract dataclass in prototype spdx 3.0
The classes `Element`, `Artifact`, `Collection` and `IntegrityMethod` should be abstract classes which implies that they can't be instantiated. It should be investigated whether we can combine dataclasses with abstract base classes and use this in our implementation.

Links that could be helpful: 
- https://stackoverflow.com/questions/60590442/abstract-dataclass-without-abstract-methods-in-python-prohibit-instantiation
- https://ivergara.github.io/ABC-and-dataclasses.html

meretp:
…d abstract. 
The [documentation for abstract methods](https://docs.python.org/3/library/abc.html#abc.abstractmethod) says
> A class that has a metaclass derived from [ABCMeta](https://docs.python.org/3/library/abc.html#abc.ABCMeta) cannot be instantiated unless all of its abstract methods and properties are overridden. 

To make the above metioned classes abstract and keep the benefits of the `dataclass_with_properties` decorator, I decided to make the `__init__` methods abstract. We implement this method for each (inherited) class anyway to use `check_types_and_set_values`. 

fixes #431 


src/spdx3/model/spdx_document.py
```diff
@@ -27,9 +28,8 @@ class SpdxDocument(Bundle):
...
     def __init__(self, spdx_id: str, creation_info: CreationInformation, name: str, elements: List[Element],
                  root_elements: List[Element], summary: Optional[str] = None, description: Optional[str] = None,
-                 comment: Optional[str] = None, verified_using: None = None, external_references: None = None,
-                 external_identifier: None = None, extension: None = None, namespace: Optional[NamespaceMap] = None,
-                 imports: Optional[List[ExternalMap]] = None, context: Optional[str] = None):
-
+                 comment: Optional[str] = None, verified_using: Optional[List[IntegrityMethod]] = None,
+                 external_references: None = None, external_identifier: None = None, extension: None = None,
+                 namespace: Optional[NamespaceMap] = None, imports: Optional[List[ExternalMap]] = None,
+                 context: Optional[str] = None):
```
armintaenzertng, line 34: `verified_using` and `imports` should default to empty lists as always

meretp: good catch

armintaenzertng (approved):
And you caught the `namespaces`, too! :D
Thanks, looks good to me now

Issue #427, opened 2023-01-17
Implement lifting SPDX2 to SPDX3


meretp:
- [ ] complete migration mapping table: 
  - [ ] add spdx 2.3 properties
  - [ ] add translation to 3.0 

- [ ] add implementation (to be decided: target- or source-based)

armintaenzertng:
Large parts of the lifting are already implemented, the Licensing is the last thing that is missing.

PR #437, opened 2023-01-19
Bump spdx2 to spdx3

meretp:
part of #427


src/spdx3/clitools/pyspdxtools3.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+@click.command()
+@click.option("--infile", "-i", prompt="input file path", help="The file containing the document to be validated or converted.")
+@click.option("--outfile", "-o", help="The file to write the converted document to (write a dash for output to stdout or omit for no conversion).")
+@click.option("--version", help='The SPDX version to be used during parsing and validation (format "SPDX-2.3").', default="SPDX-2.3")
+@click.option("--novalidation", is_flag=True, help="Don't validate the provided document.")
+def main(infile: str, outfile: str, version: str, novalidation: bool):
+    """
+    CLI-tool to parse a SPDX 2.x document and migrate it into the prototype of SPDX 3.0. As there is no definition for
```
armintaenzertng, line 32:
```suggestion
    CLI-tool to parse and validate a SPDX 2.x document and migrate it into the prototype of SPDX 3.0. As there is no definition for
```

tests/spdx3/bump/test_bump_uitls.py
```diff
@@ -0,0 +1,27 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import pytest
+
+from spdx.model.spdx_no_assertion import SpdxNoAssertion
+from spdx.model.spdx_none import SpdxNone
+from spdx3.bump_from_spdx2.bump_utils import handle_no_assertion_or_none
+
+
+@pytest.mark.parametrize("input_argument,expected_value,expected_output",
```
armintaenzertng, line 18:
```suggestion
@pytest.mark.parametrize("input_argument,expected_value,expected_stdout",
```

tests/spdx3/bump/test_file_bump.py
```diff
@@ -0,0 +1,28 @@
...
+
+@mock.patch("spdx3.model.creation_information.CreationInformation", autospec=True)
+def test_bump_file(creation_information):
+    spdx2_file: Spdx2_File = file_fixture()
+    integrity_method: Hash = Hash(HashAlgorithm.SHA1, "71c4025dd9897b364f3ebbb42c484ff43d00791c")
+
+    file: File = bump_file(spdx2_file, creation_information=creation_information)
+
+    assert file.spdx_id == "SPDXRef-File"
+    assert file.verified_using == [integrity_method]
```
armintaenzertng, line 28: What about the rest of the file properties (e.g. comment)?

armintaenzertng: see also Snippet and Package and Document below

tests/spdx3/bump/test_spdx_document_bump.py
```diff
@@ -0,0 +1,30 @@
...
+from spdx3.writer.console.spdx_id_map_writer import write_spdx_id_map
+from tests.spdx.fixtures import document_fixture
+
+
+def test_bump_spdx_document():
+    spdx2_document: Spdx2_Document = document_fixture()
+
+    spdx_id_map: SpdxIdMap = bump_spdx_document(spdx2_document)
+
+    write_spdx_id_map(spdx_id_map, sys.stdout)
```
armintaenzertng, line 26: This just tests that no errors are raised?

armintaenzertng (commented):
Solid piece of work, thanks :)

pyproject.toml
```diff
@@ -32,6 +32,7 @@ test = ["pytest"]
 
 [project.scripts]
 pyspdxtools = "spdx.clitools.pyspdxtools:main"
+pyspdxtools3 = "spdx3.clitools.pyspdxtools3:main"
```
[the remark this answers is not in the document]
meretp: Yes, I also thought about this. But as this would possibly require another restructuring (put `clitools.py` into `common`),  I decided to leave this for later.

src/spdx3/bump_from_spdx2/bump_utils.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def handle_no_assertion_or_none(field: Any, field_name: str) -> Optional[str]:
+    if isinstance(field, SpdxNone):
+        print(f"{field_name}: Missing conversion for SpdxNone.")
+        return None
+    if isinstance(field, SpdxNoAssertion):
+        return None
+    if isinstance(field, str):
+        return field
```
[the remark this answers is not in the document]
meretp: I would rather stay with the current usecase here. As soon as we need the function for other usecases (types), we can then also adapt it. I would like to avoid a kind of "to-come" implementation.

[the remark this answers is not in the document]
armintaenzertng: In that case the type hint of `field` should state `Union[SpdxNone, SpdxNoAssertion, str]` as other inputs will not return anything

tests/spdx3/bump/test_spdx_document_bump.py
meretp: Yes, this was more a temporary output. With the new added clitool we can delete this line of code and only check for the number of elements here.

tests/spdx3/bump/test_file_bump.py
meretp: The test coverage is not that good here, I agree. We would also need more tests for other "bump"-methods. As this is a rather fluent development based on the current definitions, do you think it's worth to add all these tests now?

tests/spdx3/bump/test_spdx_document_bump.py
armintaenzertng: nah, you're right. I don't think it is worth going too deep into testing before the complete migration process is done.

tests/spdx3/bump/test_bump_uitls.py
meretp: done

src/spdx3/clitools/pyspdxtools3.py
meretp: done

armintaenzertng (approved):

PR #439, opened 2023-01-24
Validate ExternalPackageRef

### LINKED ISSUE #373: Implement External Package Reference validation
part of #307

This is probably the most involved part of the validation process, earning it [an own annex](https://spdx.github.io/spdx-spec/v2.3-RC1/external-repository-identifiers/) in the specification.

Also add tests for this

armintaenzertng:
fixes #373


tests/spdx/validation/test_external_package_ref_validator.py
```diff
@@ -13,25 +13,147 @@
...
+
+CPE22TYPE_REGEX = r'^c[pP][eE]:/[AHOaho]?(:[A-Za-z0-9._\-~%]*){0,6}$'
+CPE23TYPE_REGEX = r'^cpe:2\.3:[aho\*\-](:(((\?*|\*?)([a-zA-Z0-9\-\._]|(\\[\\\*\?!"#$$%&\'\(\)\+,\/:;<=>@\[\]\^`\{\|}~]))+(\?*|\*?))|[\*\-])){5}(:(([a-zA-Z]{2,3}(-([a-zA-Z]{2}|[0-9]{3}))?)|[\*\-]))(:(((\?*|\*?)([a-zA-Z0-9\-\._]|(\\[\\\*\?!"#$$%&\'\(\)\+,\/:;<=>@\[\]\^`\{\|}~]))+(\?*|\*?))|[\*\-])){4}$'
+MAVEN_CENTRAL_REGEX = r'^[^:]+:[^:]+(:[^:]+)?$'
+NPM_REGEX = r'^[^@]+@[^@]+$'
+NUGET_REGEX = r'^[^/]+/[^/]+$'
+BOWER_REGEX = r'^[^#]+#[^#]+$'
+PURL_REGEX = r'^pkg:.+(\/.+)?\/.+(@.+)?(\?.+)?(#.+)?$'
+SWH_REGEX = r'^swh:1:(snp|rel|rev|dir|cnt):[0-9a-fA-F]{40}$'
+GITOID_REGEX = r'^gitoid:(blob|tree|commit|tag):(sha1:[0-9a-fA-F]{40}|sha256:[0-9a-fA-F]{64})$'
```
meretp, line 115: I would import these regex statements from `external_package_ref_validator.py` to prevent possible mismatch in the future if one of the regex needs changes and simply to avoid duplications.

src/spdx/validation/external_package_ref_validator.py
```diff
@@ -25,5 +36,76 @@ def validate_external_package_refs(external_package_refs: List[ExternalPackageRe
...
+        if reference_type == "maven-central":
+            return validate_against_regex(locator, MAVEN_CENTRAL_REGEX, "maven-central", context)
+        if reference_type == "npm":
+            return validate_against_regex(locator, NPM_REGEX, "npm", context)
+        if reference_type == "nuget":
+            return validate_against_regex(locator, NUGET_REGEX, "nuget", context)
+        if reference_type == "bower":
+            return validate_against_regex(locator, BOWER_REGEX, "bower", context)
+        if reference_type == "purl":
+            return validate_against_regex(locator, PURL_REGEX, "purl", context)
```
meretp, line 78: This is a lot of code duplication. We could use a dictionary instead to map `ExternalPackageRefCategory` to a set of valid types and map the regex expressions to the corresponding type. 
As I am currently working on the rdf writer and there I need to check if the specified type is listed (so one of `cpe22Type`, `cpe23Type`, `swid` and so on) or not, I would need such a mapping as well.

meretp (changes requested):
Thanks for the addition! I didn't look at each testcase and regex in detail, but overall this looks very good to me. Only two remarks from my side. 
I'm open to discuss if a dictionary approach is suitable here but as I have another use case with the rdf writer, I think it makes sense to use such a mapping.

tests/spdx/validation/test_external_package_ref_validator.py
armintaenzertng: done

src/spdx/validation/external_package_ref_validator.py
armintaenzertng: Very good point, the code is now much more concise. Please have a look if you can work with this in rdf now, too! :)

meretp (approved):
Looks very good to me and yes, I can use it now, thanks a lot! :blush:

PR #440, opened 2023-01-24
do some cleanup

maxhbr:
Just some minor cleanup of stderr vs stdout.

### DIFF
--- a/src/spdx3/clitools/pyspdxtools3.py
+++ b/src/spdx3/clitools/pyspdxtools3.py
@@ -42,12 +42,12 @@ def main(infile: str, outfile: str, version: str, novalidation: bool):
         if not novalidation:
             validation_messages: List[ValidationMessage] = validate_full_spdx_document(document, version)
             if validation_messages:
-                print("The document is invalid. The following issues have been found:")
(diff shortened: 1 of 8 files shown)


src/spdx3/clitools/pyspdxtools3.py
```diff
@@ -42,12 +42,12 @@ def main(infile: str, outfile: str, version: str, novalidation: bool):
...
             if validation_messages:
-                print("The document is invalid. The following issues have been found:")
+                print("The document is invalid. The following issues have been found:", file=sys.stderr)
                 for message in validation_messages:
-                    print(message.validation_message)
+                    print(message.validation_message, file=sys.stderr)
                 sys.exit(1)
             else:
-                print("The document is valid.")
+                print("The document is valid.", file=sys.stderr)
```
armintaenzertng, line 50: Do we want to print to `stderr` even when there are no errors?

meretp:
As agreed in the meeting today I will merge this as is. 
Regarding @armintaenzertng's comment: We have an open issue #441 to add a proper logging framework to the tools. This will also need some more research about which approach suits best and where to print which messages.

PR #446, opened 2023-01-25
[issue-442] change CLI to read version from document

### LINKED ISSUE #442: CLI: read version from document if not provided
Currently it defaults to "SPDX-2.3" and raises a validation error if the document is of another version.

armintaenzertng:
also print validation output to stderr

fixes #442

### DIFF
--- a/src/spdx/clitools/pyspdxtools.py
+++ b/src/spdx/clitools/pyspdxtools.py
@@ -26,7 +26,7 @@ from spdx.writer.write_anything import write_file
 @click.command()
 @click.option("--infile", "-i", prompt="input file path", help="The file containing the document to be validated or converted.")
 @click.option("--outfile", "-o", help="The file to write the converted document to (write a dash for output to stdout or omit for no conversion).")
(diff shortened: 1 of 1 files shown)


src/spdx/clitools/pyspdxtools.py
```diff
@@ -26,7 +26,7 @@
 @click.command()
 @click.option("--infile", "-i", prompt="input file path", help="The file containing the document to be validated or converted.")
```
meretp, line 27: Could you please also delete the prompt? I think we can do both of these changes within one PR that touches CLI.

meretp (approved):
Looks good to me, only one (resp. two) additions from my side. 

Looking at the changes I was wondering what happens if the parsed document doesn't have a version. From our implementation the parser would throw a `SPDXParsingError` whenever required fields are missing and maybe we also want to catch this to not print the whole stacktrace. What do you think?

src/spdx/clitools/pyspdxtools.py
armintaenzertng: deleted

armintaenzertng:
SPDXParsingErrors are excepted now :)

meretp:
Perfect, thanks!

Issue #451, opened 2023-01-28
Issues with DocumentNamespace and Checksum validation


anthonyharrison:
Attempting to validate a SPDX tag value results in the following errors

```
Invalid DocumentNamespace value dns:softwareassuranceguardian.com, must contain a scheme (e.g. "https:") and should not contain the "#" delimiter, line:5
VerificationCode doesn't match verifcode form, line:16
```

The namespace is compatible with the RFC and the error message does not appear to be correct.

The issue with the PackageVerificationCode:is becuase the SHA1 checkum isn't in lowercase although it does seem strange that the PackageChecksum value is valid with the checkum being valid in uppercase.

Test data is

```
SPDXVersion: SPDX-2.2
DataLicense: CC0-1.0
SPDXID: SPDXRef-DOCUMENT
DocumentName: SAG-PM generated SBOM
DocumentNamespace: dns:softwareassuranceguardian.com
Creator: Organization: dns:reliableenergyanalytics.com
Creator: Tool: SAG-PM Version: 1.2
Created: 2022-11-26T18:45:28Z
PackageName: apache-tomcat-9.0.69.zip
PackageVersion: 9.0.69
SPDXID: SPDXRef-Package-fc4a1bf0-78a0-43ca-b4a9-78adfb42138c
PackageSupplier: Organization: Apache Foundation
PackageDownloadLocation: https://dlcdn.apache.org/tomcat/tomcat-9/v9.0.69/bin/apache-tomcat-9.0.69.zip/
FilesAnalyzed: true
PackageLicenseInfoFromFiles: NOASSERTION
PackageVerificationCode: CE9F343C4BA371746FD7EAD9B59031AE34D8AFC4
PackageLicenseConcluded: NOASSERTION
PackageLicenseDeclared: NOASSERTION
PackageCopyrightText: NOASSERTION
PackageChecksum: SHA1: CE9F343C4BA371746FD7EAD9B59031AE34D8AFC4
LicenseID: LicenseRef-Unlicense
LicenseName: Unlicense
```

maxhbr:
Which version of the tools-python are you using?

anthonyharrison:
The latest one.

maxhbr:
Hey Anthony, we will have a look.

**regarding lowercase vs. uppercase:**
[...]

PR #453, opened 2023-01-30
[issue-451] use uritools.isabsuri for namespace validation

armintaenzertng:
part of #451

### DIFF
--- a/spdx/parsers/validations.py
+++ b/spdx/parsers/validations.py
@@ -34,10 +34,7 @@ def validate_is_free_form_text_or_str(value, optional=False) -> bool:
 def validate_tool_name(value, optional=False):
     striped_value = value.strip()
     if optional:
-        if len(striped_value) == 0:
-            return True
-        else:
-            return False
+        return len(striped_value) == 0
     else:
         return not (len(striped_value) == 0)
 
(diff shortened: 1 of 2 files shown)


spdx/parsers/validations.py
```diff
@@ -151,11 +152,7 @@ def validate_doc_spdx_id(value, optional=False):
...
         return optional
-    elif (
-        value.startswith("http://")
-        or value.startswith("https://")
-        or value.startswith("ftp://")
-    ) and ("#" not in value):
+    elif uritools.isabsuri(value) and ("#" not in value):
         return True
     else:
         return False
```
meretp, line 158:
```suggestion
    return uritools.isabsuri(value) and ("#" not in value)
```

I think we could shorten this a bit.

meretp (changes requested):
Thanks! One small remark from my side.

spdx/parsers/validations.py
armintaenzertng: Good point. I was reluctant to touch the structure at first, but going through the whole file erasing lines was quite satisfying :D

meretp (approved):
Looks good to me, I didn't intend to change the whole file, but this clearly improves the readability, thank you!

PR #461, opened 2023-02-01
[issue-386] Finish Package validation

### LINKED ISSUE #386: Finish package validation
part of #307 

- [x] check if verification code is required when `files_analyzed = True` and make check for it
- [x] write tests for package verification code validation
- [x] write tests for "no `CONTAINS` or `CONTAINED_BY` relationships if `files_analyzed = False`"

armintaenzertng:
fixes #386


src/spdx/validation/package_validator.py
```diff
@@ -43,23 +44,18 @@ def validate_package_within_document(package: Package, document: Document) -> Li
...
+        combined_relationships: List[Relationship] = package_contains_relationships + contained_in_package_relationships
 
-        contained_in_package_relationships = [relationship for relationship in document.relationships if
-                                              relationship.relationship_type == RelationshipType.CONTAINED_BY and relationship.related_spdx_element_id == package.spdx_id]
-        if contained_in_package_relationships:
+        if combined_relationships:
             validation_messages.append(
                 ValidationMessage(
-                    f"package must contain no elements if files_analyzed is False, but found {contained_in_package_relationships}",
+                    f"package must contain no elements if files_analyzed is False, but found {combined_relationships}",
```
meretp, line 58: Reading this, I was wondering if a package must contain no elements in general or only files? The spec is mostly referring to files and I am not sure, if a package could contain anything other than a file.

meretp (approved):
Thanks! Two small remarks from my side, not really blocking, so I would approve this as is.

src/spdx/validation/package_validator.py
armintaenzertng: I don't think it can contain anything else than files. This is not explicitly stated by the spec for relationships, so who knows... I would leave it like that for now.

tests/spdx/validation/test_package_verification_code_validator.py
```diff
@@ -0,0 +1,38 @@
...
+
+    assert validation_messages == []
+
+
+@pytest.mark.parametrize("code, expected_message",
+                         [(PackageVerificationCode("71c4025dd9897b364f3ebbb42c484ff43d00791cab", []),
+                           "value of verification_code must consist of 40 lowercase hexadecimal digits, but is: 71c4025dd9897b364f3ebbb42c484ff43d00791cab (length: 42 digits)"),
+                          (PackageVerificationCode("71c4025dd9897b364f3ebbb42c484ff43d00791c", ["/invalid/excluded/file"]),
+                           'file name must not be an absolute path starting with "/", but is: /invalid/excluded/file')
+                          ])
```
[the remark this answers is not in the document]
armintaenzertng: ah yes, and I wanted to give checksums the same treatment

PR #467, opened 2023-02-08
[issue-466] bug fix and add test

### LINKED ISSUE #466: Tag-value writer: pacakges should precede contained files
According to the [spec](https://spdx.github.io/spdx-spec/v2.3/composition-of-an-SPDX-document/#5.2.3):

> Files are assumed to be associated with the package information that immediately precedes it, if a package exists.

[The implemented logic](https://github.com/spdx/tools-python/blob/788e8d4aec940a3c2b2f16bd731564ee1a8da5d8/src/spdx/writer/tagvalue/tagvalue_writer_helper_functions.py#L84) in the tag-value writer checks against strings and not `RelationshipType` which leads to output where contained files do not follow the related package.

meretp:
check against RelationshipType to let package information precede contained file information and add a test for the helper function 

fixes #466


maxhbr (approved):
LGTM

PR #471, opened 2023-02-09
make it possibile to set encoding

chrisdecker1201:


meretp:
Hi @chrisdecker1201, thanks for your contribution! This looks like a useful feature. Can you please add the signoff to your commits so that this can be merged? (see [here](https://github.com/spdx/tools-python/blob/main/CONTRIBUTING.md#licensing)

Also, I have one question out of curiosity: I would assume that you use the tools-python as a library and as you also spoke up for the feature to merge SPDX documents: How or in what context are you using the tools?

chrisdecker1201:
@meretp Your assumption is correct 😄. We are getting a SPDX file from a third-party application and have to add additional SPDX information to this file. Therefore our developers should checkin this additional SPDX information into the repository and we are developing a tool for combining the files in the CI/CD process.

Additionally we use this to validate the SPDX file.

requirements.txt
```diff
@@ -0,0 +1,7 @@
+click
+ply
+pytest
+pyyaml
+rdflib
+uritools
+xmltodict
```
meretp, line 7: I would prefer to avoid duplication here as we already have the dependencies in the `pyproject.toml`.

CONTRIBUTING.md
```diff
@@ -26,9 +26,12 @@ Here's the process to make changes to the codebase:
 2. Review [open pull requests](https://github.com/spdx/tools-python/pulls) before committing time to a substantial
    revision. Work along similar lines may already be in progress.
 
-3. Create a new branch:
+3. Create a new branch and set up environment:
    ```sh
    git checkout -b fix-or-improve-something
+   python -m venv ./venv
+   ./venv/bin/activate
+   pip install -r requirements.txt
```
meretp, line 34: We could simply write `pip install -e .` here and don't need an extra requirements.txt as commented below.

spdx/parsers/parse_anything.py
```diff
@@ -21,10 +21,13 @@
 from spdx.parsers.builderexceptions import FileTypeError
 
 
-def parse_file(fn):
+def parse_file(fn, encoding="utf-8"):
+    in_mode = "r"
     builder_module = jsonyamlxmlbuilders
     read_data = False
     if fn.endswith(".rdf") or fn.endswith(".rdf.xml"):
+        in_mode = "rb"
```
meretp, line 29: I don't think we need this mode as a graph can be parsed from a file-like object.

meretp (changes requested):
Thanks for adding the sign-off! I have some small remarks on the last changes.

requirements.txt
chrisdecker1201: Good to know 😄

meretp (approved):
Thanks for addressing the remarks, this looks good to me! And also thanks for sharing your usecase!

PR #473, opened 2023-02-10
Update prototype

meretp:
update SPDX3.0 prototype according to the latest changes in spdx-3-model repo

### DIFF
--- a/src/spdx3/model/annotation.py
+++ b/src/spdx3/model/annotation.py
@@ -32,12 +32,12 @@ class AnnotationType(Enum):
 @dataclass_with_properties
 class Annotation(Element):
     annotation_type: AnnotationType = None
-    subject: List[str] = field(default_factory=list)
+    subject: str = None
     content_type: Optional[str] = None  # placeholder for MediaType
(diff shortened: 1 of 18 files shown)


src/spdx3/model/annotation.py
```diff
@@ -32,12 +32,12 @@ class AnnotationType(Enum):
 @dataclass_with_properties
 class Annotation(Element):
     annotation_type: AnnotationType = None
-    subject: List[str] = field(default_factory=list)
+    subject: str = None
```
armintaenzertng, line 36: `annotation_type` and `subject` are required fields that should not have a default value

armintaenzertng (approved):
Thanks, looks good :)

src/spdx3/model/annotation.py
meretp: Agreed but `Annotation` is a child class of `Element` which already has optional fields. So we need to set default values here. But the implementation of the `__init__` function takes care of these fields and doesn't set default values for the  corresponding parameters.

Issue #464, opened 2023-02-06
Supplier name could be NOASSERTION


spatil00:
Supplier name could be NOASSERTION , if this is not added then code throws exception. 

SPDX specification 2.3  says for supplier tag in package. 

"Use NOASSERTION if:

the SPDX document creator has attempted to but cannot reach a reasonable objective determination;

the SPDX document creator has made no attempt to determine this field; or

the SPDX document creator has intentionally provided no information (no meaning should be implied by doing so)."

Check needs to be added to the rdfbuilder.py in function EntityBuilder

spatil00:
I have fix , how can I contribute to fix this issue ?

meretp:
Hi @spatil00! 
Thanks for the report and the fix, please open a PR for the fix and I will have a look! :)

spatil00:
Hi , I don't have access to create branch , could you please help with that.

meretp:
Sure. You need to fork the repo first (this seems to be missing in the contributing guideline), within your own repository you can create a new branch from main and commit your changes to fix the issue. You can then create a pull request from your branch (in your fork) to the original main branch. 
For more information about forks, see here: https://docs.github.com/en/get-started/quickstart/fork-a-repo 
I hope, this helps. 
I will open an issue to add the fork step to the contributing guideline. Sorry for the circumstances here!

spatil00:
Hi @meretp ,  I have created pull request , I am not able to assign that to you , for review but I think you can help me with further process now. 

https://github.com/spdx/tools-python/pull/475

Thanks & Regards
Sandeep

PR #475, opened 2023-02-13
Issue #464 Fix: Added check for NOASSERTION 

spatil00:
Issue #464 Fix for issue

### DIFF
--- a/spdx/parsers/jsonyamlxml.py
+++ b/spdx/parsers/jsonyamlxml.py
@@ -1244,8 +1244,9 @@ class PackageParser(BaseParser):
         """
         if isinstance(pkg_supplier, str):
             if pkg_supplier == "NOASSERTION":
-                return self.builder.set_pkg_supplier(self.document, NoAssert())
-            entity = self.builder.create_entity(self.document, pkg_supplier)
+                return self.builder.set_pkg_supplier(self.document,utils.NoAssert())
+            else:
+                entity = self.builder.create_entity(self.document, pkg_supplier)
             try:
                 return self.builder.set_pkg_supplier(self.document, entity)
             except SPDXValueError:
@@ -1264,8 +1265,9 @@ class PackageParser(BaseParser):
         """
         if isinstance(pkg_originator, str):
             if pkg_originator == "NOASSERTION":
-                return self.builder.set_pkg_originator(self.document, NoAssert())
-            entity = self.builder.create_entity(self.document, pkg_originator)
+                return self.builder.set_pkg_originator(self.document,utils.NoAssert())
+            else:
+                entity = self.builder.create_entity(self.document, pkg_originator)
             try:
                 return self.builder.set_pkg_originator(self.document, entity)
             except SPDXValueError:
--- a/spdx/parsers/rdfbuilders.py
+++ b/spdx/parsers/rdfbuilders.py
@@ -16,6 +16,7 @@ from spdx import file
 from spdx import license
 from spdx import package
 from spdx import version
+from spdx import utils
 from spdx.checksum import Checksum, ChecksumAlgorithm
 from spdx.document import Document
 from spdx.parsers.builderexceptions import CardinalityError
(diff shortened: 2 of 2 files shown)


spdx/parsers/rdfbuilders.py
```diff
@@ -163,6 +164,8 @@ def create_entity(self, doc, value):
             return self.build_person(doc, value)
         elif self.org_re.match(value):
             return self.build_org(doc, value)
+        elif value == "NOASSERTION":            # Supplier name could be NOASSERTION SPDX 2.3
+            return  utils.NoAssert()
```
meretp, line 168: If you in include this check here, any field that is created through this method, so basically every entity (e.g. the creator of the document, any reviewer, any annotator) could also be set to `NoAssert` which is not valid in general.

meretp (changes requested):
Thanks for your contribution, the current codebase is not that easy to understand and your fix basically does what it should (especially it would also fix the same problem with `package originator` which could also be `NOASSERTION`). But your current solution would also allow other fields to be `NOASSERTION` which is not valid per spec (see my comment below). 
Looking deeper in the code, I have seen that the case distinction to `NOASSERTION` is already made [here](https://github.com/spdx/tools-python/blob/d6932e2d295a53c619deffd949ec30622a248765/spdx/parsers/rdf.py#L578) (and [here](https://github.com/spdx/tools-python/blob/d6932e2d295a53c619deffd949ec30622a248765/spdx/parsers/rdf.py#L564) for the package originator), just currently not seem to work properly. Could you try to fix the issue there?

spatil00:
@meretp Thanks for comments , I have moved fix into jsonyamlxml.py , Its specific to supplier tag now, also I have taken liberty to fix same issue for package originator tag. Please let me know if this fix is fine can be merged.

meretp:
Thanks for the effort, but I am sorry, maybe I wasn't clear on that, this issue is already fixed in jsonyamlxml.py. 
You could easily check this if you try to parse a Json document with supplier "NOASSERTION" on main. Your current fix would still fail for a RDF-file with package supplier being "NOASSERTION". 

You need to adapt the code in `spdx/parser/rdf.py` as I tried to point out. The following [logic](https://github.com/spdx/tools-python/blob/d6932e2d295a53c619deffd949ec30622a248765/spdx/parsers/rdf.py#L578) needs to be adapted:
```
 def p_pkg_suppl(self, p_term, predicate):
        for _s, _p, o in self.graph.triples((p_term, predicate, None)):
            try:
                if o == "NOASSERTION":
                    self.builder.set_pkg_supplier(self.doc, utils.NoAssert())
```
The return value `o` of `graph.triples()` is a `rdflib.Literal` and the comparison `o == "NOASSERTION"` is always False, so one would need to cast the type to `str` for this comparison. The same holds for `p_pkg_originator` some lines higher.

spdx/parsers/rdfbuilders.py
```diff
@@ -162,7 +162,7 @@ def create_entity(self, doc, value):
         elif self.person_re.match(value):
             return self.build_person(doc, value)
         elif self.org_re.match(value):
-            return self.build_org(doc, value)
+            return self.build_org(doc, value)
```
DjDeepDive, line 165: No need to commit random whitespace, please

spdx/parsers/jsonyamlxml.py
```diff
@@ -1264,8 +1265,9 @@ def parse_pkg_originator(self, pkg_originator):
         """
         if isinstance(pkg_originator, str):
             if pkg_originator == "NOASSERTION":
-                return self.builder.set_pkg_originator(self.document, NoAssert())
-            entity = self.builder.create_entity(self.document, pkg_originator)
+                return self.builder.set_pkg_originator(self.document,utils.NoAssert())
```
DjDeepDive, line 1268: Stylistic check: Can you put a space between the comma and "utils" to be consistent on formatting?

meretp:
@spatil00 we want to make a fix release soon where this issue should also be fixed. Do you plan to address the requested changes soon?

spatil00:
Looks like the issue that I wanted to fix was already got fixed as mentioned by @meretp,

PR #480, opened 2023-02-14
[issue-479] fix parsing of snippet ranges

### LINKED ISSUE #479: PackageVersion with ":" are considered invalid token
While applying [sbomqs quality checks](https://github.com/interlynk-io/sbomqs) on SBOMs, we found the parser failing to parse versions with ":" in them

```
 pyspdxtools_parser --file bom.nginx.spdx
```

results in the attached file with the missing field -

```
PackageVersion must be single line of text, line: 148
PackageVersion must be single line of text, line: 272
...
```

However, the included versions is indeed 

```
PackageName: bsdutils
PackageVersion: 1**:**2.36.1-8+deb11u1
```

This is not an issue for spdx-json because of the quotes.

**What did you expect to happen?**
Accept the valid versions correctly.

**What happened instead?**
[bom.nginx.spdx.txt](https://github.com/spdx/tools-python/files/10726753/bom.nginx.spdx.txt)

parser considers the SBOM invalid

**Additional details (base image name, container registry info...):**

meretp:
The previous regular expressions for ranges would prevent lines with ":" to be parsed as LINE and instead be parsed as RANGE, the remainder after ":\d+" would be truncated and ignored, this lead to problems when parsing a package version like "1:2.36.1-8+deb11u1".

fixes #479

### DIFF
--- a/spdx/parsers/lexers/tagvalue.py
+++ b/spdx/parsers/lexers/tagvalue.py
@@ -8,6 +8,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import re
 
 from ply import lex
 
@@ -170,10 +171,6 @@ class Lexer(object):
         t.value = t.value[1:].strip()
         return t
 
-    def t_RANGE(self, t):
-        r":\s*\d+:\d+"
-        t.value = t.value[1:].strip()
-        return t
 
     def t_DOC_REF_ID(self, t):
         r":\s*DocumentRef-([A-Za-z0-9\+\.\-]+)"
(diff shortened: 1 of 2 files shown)


maxhbr (approved):
LGTM

spdx/parsers/lexers/tagvalue.py
```diff
@@ -171,7 +171,7 @@ def t_CHKSUM(self, t):
         return t
 
     def t_RANGE(self, t):
-        r":\s*\d+:\d+"
+        r":\s*\d+:\d+[^A-Za-z\n.]"
```
maxhbr, line 174: this no longer matches regular expressions with only one digit after `:`

maxhbr (changes requested):

maxhbr (approved):
LGTM

maxhbr (approved):
LGTM

Issue #463, opened 2023-02-02
Implement SPDX2.2 specific validation


armintaenzertng:
Currently, only SPDX2.3 validation is implemented. 2.2 has more fields marked as mandatory and a few fields don't exist in that version.

armintaenzertng:
fixed by #490

Issue #360, opened 2022-12-08
Add examples using the new data model


nicoweidner:
We have to adapt the examples to work with the refactored data model

PR #482, opened 2023-02-14
Clean up examples

meretp:
As a start to solving #360 I deleted those files which aren't valid according to the spec. Or more precise: json, yaml and xml formats weren't defined in specVersion 2.1, so it does not make sense to keep these examples in the repo. Additionally, I deleted some data in the SBOM file which were added later on, probably to test a feature development but doesn't make sense for the original file. Trying to parse the SBOM file would still cause a ValidationError but as this is a file from specVersion 2.2 and the validation is still an open issue (#463), I think we can leave the example as is in the repo.


maxhbr (approved):
LGTM

PR #484, opened 2023-02-15
fix helper method

meretp:
The return value of err.args[0] is a string, so we need to use append to add this string to the list of messages in the logger.


maxhbr (approved):
LGTM

PR #485, opened 2023-02-15
add GitHub Actions workflow

meretp:
With this PR we add a GitHub Actions workflow to the main branch to also test installation on windows.


meretp:
To properly test the installation on windows, we shouldn't use bash as shell in the github actions but probably power shell. Converting this to a draft as it will need more work.

armintaenzertng (approved):
I think we can remove CircleCI now, but this looks good for now :)

PR #486, opened 2023-02-15
use datetime.fromisoformat to parse date string, instead of own imple…

chrisdecker1201:
I tried to validate a date in the following format and got an error

`2023-02-15T11:13:01.729997`

This is the output from: `datetime.utcnow().isoformat()` (https://docs.python.org/3/library/datetime.html#datetime.date.isoformat)

As you describe in the comment that you want an ISO 8601 format I think it would be better to use the python default method to validate the date

`datetime.fromisoformat(string)`. (https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat)

Therefore the above datetime is a valid datetime with no timezone

From the python documentation:         
The full format looks like 'YYYY-MM-DD HH:MM:SS.mmmmmm'. By default, the fractional part is omitted if self.microsecond == 0.
If self.tzinfo is not None, the UTC offset is also attached, giving a full format of 'YYYY-MM-DD HH:MM:SS.mmmmmm+HH:MM'.

### DIFF
--- a/spdx/utils.py
+++ b/spdx/utils.py
@@ -9,9 +9,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import datetime
 import hashlib
 import re
+from datetime import datetime
 from typing import Dict, List, TYPE_CHECKING
 
 from ply import lex
@@ -30,21 +30,12 @@ def datetime_iso_format(date):
     """
     Return an ISO-8601 representation of a datetime object.
     """
-    return date.isoformat() + "Z"
-
+    return date.strftime('%Y-%m-%dT%H:%M:%SZ')
 
 # Matches an iso 8601 date representation
 DATE_ISO_REGEX = re.compile(
-    r"(\d\d\d\d)-(\d\d)-(\d\d)T(\d\d):(\d\d):(\d\d)Z", re.UNICODE
+    r"(\d{4})-(\d{2})-(\d{2})T(\d{2}):(\d{2}):(\d{2})Z", re.UNICODE
 )
-
-# Groups for retrieving values from DATE_ISO_REGEX matches.
-DATE_ISO_YEAR_GRP = 1
-DATE_ISO_MONTH_GRP = 2
-DATE_ISO_DAY_GRP = 3
-DATE_ISO_HOUR_GRP = 4
-DATE_ISO_MIN_GRP = 5
-DATE_ISO_SEC_GRP = 6
 
 
 def datetime_from_iso_format(string):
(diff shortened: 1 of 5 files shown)


chrisdecker1201:
I'm confused about the build output because locally all tests run through.

meretp:
When I fetch this branch the tests also fail locally, maybe you need to run `pip install .` on your current changes?

meretp:
I guess the "problem" is your python version, the documentation fpr `datetime.fromisoformat` states: 

> Changed in version 3.11: Previously, this method only supported the format YYYY-MM-DD.

So we probably can't this method as we also support older Python versions.

chrisdecker1201:
Maybe this a solution for you now?

zvr:
I don't know what field this is about, but please note that SPDX spec does *not* allow fractional seconds.

Only `YYYY-MM-DDThh:mm:ssZ` is allowed for [Created](https://spdx.github.io/spdx-spec/v2.3/document-creation-information/#69-created-field), for example. The `Z` is mandatory.

Therefore the example in the first comment here is invalid.

chrisdecker1201:
@zvr Oh thank you :). I expected ISO 8601 because this is mentioned in the code comments.

spdx/utils.py
```diff
@@ -30,42 +30,27 @@ def datetime_iso_format(date):
...
-            hour=int(match.group(DATE_ISO_HOUR_GRP)),
-            second=int(match.group(DATE_ISO_SEC_GRP)),
-            minute=int(match.group(DATE_ISO_MIN_GRP)),
-        )
-        return date
-    else:
+    if not match:
         return None
 
+    if string.endswith("Z"):
```
zvr, line 50: The `Z` is mandatory, so the `if` will always succeed.

zvr:
In general, I am against this change: `fromisoformat()`, regardless of its name, does not handle ISO formats (e.g. the `Z` suffix). And SPDX does not accept any ISO anyway.

Therefore the custom parsing of the specific string allowed by SPDX is the cleaner solution.

chrisdecker1201:
@zvr I'm totally with you.

meretp:
One more general comment from my side: We are currently refactoring the whole codebase and hope to get the new refactored version out soon. It's hard to really name a timeline but I would hope that the refactored version is out within the next month (without any guarantee). The refactored version will follow the spec very close and will have separated layers for parsing, writing and validating SPDX documents. We don't really change this particular function, so we are not using `fromisoformat()` but this should be a more general hint about changes to the current code which aren't fixes. 
You find the current state of development on the [refactored-python-tools](https://github.com/spdx/tools-python/tree/refactor-python-tools) branch. Since you seem to be adding the `tools-python` to your code, it might be worth it to have a look at this new version and decide if you want to adapt to the new version. From your comment on the other PR I guess you need the version on pypi to really use it, right? We will try to get this out as soon as possible.

chrisdecker1201:
@meretp Yes I need a version to really use it :)

PR #493, opened 2023-02-23
[issue-374] implement license expression validation

### LINKED ISSUE #374: Implement license expression validation
can be done after #10 
part of #307 

Also add more tests for this

armintaenzertng:
fixes #374

### DIFF
--- a/src/spdx/validation/license_expression_validator.py
+++ b/src/spdx/validation/license_expression_validator.py
@@ -38,7 +38,7 @@ def validate_license_expression(license_expression: Optional[
     if license_expression in [SpdxNoAssertion(), SpdxNone(), None]:
         return []
 
-    if context is None:
+    if not context:
         context = ValidationContext(parent_id=parent_id, element_type=SpdxElementType.LICENSE_EXPRESSION, full_element=license_expression)
 
     validation_messages = []
@@ -55,14 +55,16 @@ def validate_license_expression(license_expression: Optional[
     try:
         get_spdx_licensing().parse(str(license_expression), validate=True, strict=True)
     except ExpressionParseError as err:
+        # This error is raised when an exception symbol is used as a license symbol and vice versa.
+        # So far, it only catches the first such error in the provided string.
         validation_messages.append(
             ValidationMessage(
                 f"{err}. for license_expression: {license_expression}",
                 context)
         )
     except ExpressionError:
-        # This happens for invalid symbols but the error provides only a string of these. On the other hand,
-        # get_spdx_licensing().validate() gives an actual list of invalid symbols, so this is handled above.
+        # This error is raised for invalid symbols within the license_expression, but it provides only a string of these.
+        # On the other hand, get_spdx_licensing().validate() gives an actual list of invalid symbols, so this is handled above.
         pass
 
     return validation_messages
--- a/src/spdx/validation/license_expression_validator.py
+++ b/src/spdx/validation/license_expression_validator.py
@@ -11,25 +11,58 @@
 
 from typing import List, Optional, Union
 
-from license_expression import LicenseExpression
+from license_expression import LicenseExpression, get_spdx_licensing, ExpressionError, ExpressionParseError
+from spdx.model.document import Document
+
 from spdx.model.spdx_no_assertion import SpdxNoAssertion
 from spdx.model.spdx_none import SpdxNone
-from spdx.validation.validation_message import ValidationMessage
+from spdx.validation.validation_message import ValidationMessage, ValidationContext, SpdxElementType
 
 
 def validate_license_expressions(license_expressions: Optional[
-    Union[List[LicenseExpression], SpdxNoAssertion, SpdxNone]]) -> List[ValidationMessage]:
+    Union[List[LicenseExpression], SpdxNoAssertion, SpdxNone]], document: Document, parent_id: str) -> List[ValidationMessage]:
     if license_expressions in [SpdxNoAssertion(), SpdxNone(), None]:
         return []
 
-    error_messages = []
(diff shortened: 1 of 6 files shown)


src/spdx/validation/license_expression_validator.py
```diff
@@ -11,25 +11,58 @@
...
+
+    validation_messages = []
+    license_ref_ids: List[str] = [license_ref.license_id for license_ref in document.extracted_licensing_info]
 
-    return error_messages
+    for non_spdx_token in get_spdx_licensing().validate(license_expression).invalid_symbols:
+        if non_spdx_token not in license_ref_ids:
+            validation_messages.append(
+                ValidationMessage(
+                    f"Unrecognized license reference: {non_spdx_token}. license_expression must only use IDs from the license list or extracted licensing info, but is: {license_expression}",
```
meretp, line 51:
```suggestion
                    f"Unrecognized license reference: {non_spdx_token}. License_expression must only use IDs from the license list or extracted licensing info, but is: {license_expression}",
```

tests/spdx/validation/test_license_expression_validator.py
```diff
@@ -10,14 +10,94 @@
...
+                           f'A plain license symbol cannot be used as an exception in a "WITH symbol" statement. for token: "{FIXTURE_LICENSE_ID}" at position: 39. for license_expression: GPL-2.0-or-later AND {FIXTURE_LICENSE_ID} WITH {FIXTURE_LICENSE_ID}'),
+                          (f"GPL-2.0-or-later with MIT and {FIXTURE_LICENSE_ID} with GPL-2.0-or-later",
+                           f'A plain license symbol cannot be used as an exception in a "WITH symbol" statement. for token: "MIT" at position: 22. for license_expression: GPL-2.0-or-later WITH MIT AND {FIXTURE_LICENSE_ID} WITH GPL-2.0-or-later'),
+                          ("389-exception with 389-exception",
+                           'A license exception symbol can only be used as an exception in a "WITH exception" statement. for token: "389-exception". for license_expression: 389-exception WITH 389-exception'),
+                          ("389-exception with MIT",
+                           'A license exception symbol can only be used as an exception in a "WITH exception" statement. for token: "389-exception". for license_expression: 389-exception WITH MIT'),
```
meretp, line 91: The validation provided by the `license_expression` package seems to return after the first error found which is not consistent with our approach. That's unfortune but probably not changeable, only to mention it here.

src/spdx/validation/license_expression_validator.py
```diff
@@ -11,25 +11,58 @@
...
+        get_spdx_licensing().parse(str(license_expression), validate=True, strict=True)
+    except ExpressionParseError as err:
+        validation_messages.append(
+            ValidationMessage(
+                f"{err}. for license_expression: {license_expression}",
+                context)
+        )
+    except ExpressionError:
+        # This happens for invalid symbols but the error provides only a string of these. On the other hand,
+        # get_spdx_licensing().validate() gives an actual list of invalid symbols, so this is handled above.
```
meretp, line 65:
```suggestion
        # This error is raised for invalid symbols within the license_expression but it provides only a string of these. On the other hand,
        # get_spdx_licensing().validate() gives an actual list of invalid symbols, so this is handled above.
```
At first I was very confused about this comment, but with the previous discussion and the discussion in the [issue](https://github.com/nexB/license-expression/issues/82) in mind it got clearer. Although I'm not a fan of comments in general, maybe you could add one for the `ExpressionParseError` that it is used to catch falsy exceptions.

src/spdx/validation/license_expression_validator.py
meretp: Seeing that the error message from the `ExpressionParseError` also uses lower case after a `.` you can ignore this remark to keep it "somehow" consistent, but I don't really like this.

meretp (approved):
Thanks for adding this! Only a few nitpicking remarks from my side that are not merge-blocking.

src/spdx/validation/license_expression_validator.py
armintaenzertng: yeah, not ideal, but it should work nevertheless ;)

src/spdx/validation/license_expression_validator.py
```diff
@@ -11,25 +11,58 @@
...
+
+    return validation_messages
+
+
+def validate_license_expression(license_expression: Optional[
+    Union[LicenseExpression, SpdxNoAssertion, SpdxNone]], document: Document, parent_id: str, context: ValidationContext = None) -> List[ValidationMessage]:
+    if license_expression in [SpdxNoAssertion(), SpdxNone(), None]:
+        return []
+
+    if context is None:
```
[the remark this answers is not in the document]
armintaenzertng: done

tests/spdx/validation/test_license_expression_validator.py
armintaenzertng: Yes, this is also not ideal, but it really is only a workaround until maybe `get_spdx_licensing().validate()` lists the full list of errors.

src/spdx/validation/license_expression_validator.py
armintaenzertng: done

src/spdx/validation/validation_message.py
```diff
@@ -25,7 +26,6 @@ class SpdxElementType(Enum):
     PACKAGE = auto()
     FILE = auto()
     SNIPPET = auto()
-    LICENSE = auto()
```
meretp, line 28: Good catch!

meretp (approved):
Thanks!

PR #494, opened 2023-02-24
[issue-441] replace print() with logging

### LINKED ISSUE #441: Add a proper logging framework
Making sure that `stdout` is not cluttered with unwanted output that could go to `stderr`

armintaenzertng:
fixes #441

### DIFF
--- a/src/spdx/clitools/pyspdxtools.py
+++ b/src/spdx/clitools/pyspdxtools.py
@@ -10,6 +10,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import logging
 import sys
 from typing import List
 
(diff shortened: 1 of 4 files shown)


meretp (changes requested):
Your changes look good to me. I found two additional places in the codebase where information should bel logged. As I can't comment on unchanged lines, I reference them here: 
- the [output to the console](https://github.com/spdx/tools-python/blob/c8fd2d7ac7547e1db1c4aef5556238fa9c430d48/src/spdx/clitools/pyspdxtools.py#L42) is a logging.info  
- the system exits [here](https://github.com/spdx/tools-python/blob/c8fd2d7ac7547e1db1c4aef5556238fa9c430d48/src/spdx/parser/rdf/creation_info_parser.py#L75-L86) should also be logged

Within the parser I also implemented a `Logger`- class, I didn't look in the documentation for the `logging` package but do you know if we could merge this somehow? So does the `logging` package provide any options for an individual logger? Although I am not sure if we would mix different levels of logging when merging this.

armintaenzertng:
I fixed the system exits' logs, but for the cli tool parser the output is the actual wanted result of the tool and should therefore go to stdout, I think.

meretp (approved):
Thanks for addressing my remark and I agree that the output is wanted as is. 
Do you have any thoughts on my last question concerning an indiviual logger?

armintaenzertng:
Concerning the parsing logger, which basically just collects all parsing errors, I think that it is still valid as it is go give helpful information for the `SpdxParsingError` that we raise (this can be caught by library users, which can easily do whatever they want with the `get_messages()` function).
The question would be if we additionally want to separately log all these errors. But right now, I don't see the merit in that.

PR #497, opened 2023-02-28
[issue-464] fix rdf parser: package supplier and originator could be …

### LINKED ISSUE #464: Supplier name could be NOASSERTION

meretp:
…NOASSERTION

fixes #464

### DIFF
--- a/spdx/parsers/rdf.py
+++ b/spdx/parsers/rdf.py
@@ -114,11 +114,11 @@ class BaseParser(object):
         NONE, NOASSERTION or UNKNOWN if so returns proper model.
         else returns value
         """
-        if value == self.spdx_namespace.none:
+        if value == self.spdx_namespace.none or str(value) == "NONE":
             return utils.SPDXNone()
-        elif value == self.spdx_namespace.noassertion:
+        elif value == self.spdx_namespace.noassertion or str(value) == "NOASSERTION":
             return utils.NoAssert()
-        elif value == self.spdx_namespace.unknown:
+        elif value == self.spdx_namespace.unknown or str(value) == "UNKNOWN":
             return utils.UnKnown()
         else:
             return str(value)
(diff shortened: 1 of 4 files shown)


spdx/parsers/rdf.py
```diff
@@ -561,10 +561,11 @@ def p_pkg_files_analyzed(self, p_term, predicate):
...
         for _s, _p, o in self.graph.triples((p_term, predicate, None)):
             try:
-                if o == "NOASSERTION":
-                    self.builder.set_pkg_originator(self.doc, utils.NoAssert())
+                originator = self.to_special_value(o)
+                if isinstance(originator, utils.NoAssert):
+                    self.builder.set_pkg_originator(self.doc, originator)
                 else:
-                    ent = self.builder.create_entity(self.doc, str(o))
+                    ent = self.builder.create_entity(self.doc, originator)
```
armintaenzertng, line 568: At this point, `originator` could be `utils.SPDXNone`, but `create_entity` expects it to be a string.

See below for `supplier`, too.

meretp: True, it could also be `utils.UnKnown`. This isn't explicitly allowed or forbidden by the spec, so I simply added these to the check for instances above. As we are not fully supporting rdf here, I think this is alright.

armintaenzertng (approved):
Thanks :)

PR #499, opened 2023-03-01
[issue-479] fix parsing of package version that looks like a range value

meretp:
To be able to also parse package versions that might look like a range, e.g. `1:2` I decided to delete the token `RANGE` from the tag-value parser. The build functions for `SnippetByteRange` and `SnippetLineRange` therefore need to check if the parsed value matches the expected pattern and will raise a `SPDXValueError` if not. Since the tests in `test_builder.py` are wrapped in a class that is inherited from `TestCase` I can't use `pytest.mark.parametrize` and additional parameters for the tests `test_snippet_byte_range_value` and `test_snippet_line_range_value`.

### DIFF
--- a/spdx/parsers/lexers/tagvalue.py
+++ b/spdx/parsers/lexers/tagvalue.py
@@ -137,7 +137,6 @@ class Lexer(object):
         "PERSON_VALUE",
         "DATE",
         "LINE",
-        "RANGE",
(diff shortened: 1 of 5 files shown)


armintaenzertng (approved):
Looks fine, thanks :)
GitHub Packages is currently down, that's why the CI fails.

PR #501, opened 2023-03-02
[fix] use correct value for LicenseListVersion

meretp:
While working on the parser  for tag-value files I came across this little bug.


armintaenzertng (approved):
good catch :)

PR #510, opened 2023-03-07
fix creation info writer test

meretp:
Working on the tag-value parser I got a warning from pytest that lead me to this small bug.


armintaenzertng (approved):
good catch! :)

PR #517, opened 2023-03-10
Fix tag value writer

meretp:
Due to the type conversion before entering the `set_value` method a `None` value for `LicenseListVersion` was written to the outptut file. The `f-string` calls the implemented `str`-conversion for the instance of `Version`, so we don't need to do this conversion explicitly. 

To improve the test coverage here, I also added a test.

### DIFF
--- a/tests/spdx/writer/tagvalue/test_creation_info_writer.py
+++ b/tests/spdx/writer/tagvalue/test_creation_info_writer.py
@@ -1,0 +1,44 @@
+# SPDX-FileCopyrightText: 2023 SPDX Contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+import datetime
+from unittest.mock import mock_open, patch, call, MagicMock
+
+import pytest
+
+from spdx.model.document import CreationInfo
+from tests.spdx.fixtures import creation_info_fixture, actor_fixture
+
+from spdx.writer.tagvalue.creation_info_writer import write_creation_info
+
+
+@pytest.mark.parametrize("creation_info, expected_calls",
+                         [(creation_info_fixture(), [call("SPDXVersion: SPDX-2.3\n"), call("DataLicense: CC0-1.0\n"),
+                                                     call("SPDXID: SPDXRef-DOCUMENT\n"),
+                                                     call("DocumentName: documentName\n"),
+                                                     call("DocumentNamespace: https://some.namespace\n"),
+                                                     call("DocumentComment: documentComment\n"),
+                                                     call("\n## External Document References\n"), call(
+                                 "ExternalDocumentRef: DocumentRef-external https://namespace.com SHA1: 71c4025dd9897b364f3ebbb42c484ff43d00791c\n"),
+                                                     call("\n"), call("## Creation Information\n"),
+                                                     call("LicenseListVersion: 3.19\n"),
+                                                     call("Creator: Person: creatorName (some@mail.com)\n"),
+                                                     call("Created: 2022-12-01T00:00:00Z\n"),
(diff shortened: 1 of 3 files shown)


tests/spdx/writer/tagvalue/test_creation_info_writer.py
```diff
@@ -0,0 +1,45 @@
...
+                           [call("SPDXVersion: SPDX-2.3\n"), call("DataLicense: CC0-1.0\n"),
+                            call("SPDXID: SPDXRef-DOCUMENT\n"), call("DocumentName: Test document\n"),
+                            call("DocumentNamespace: https://namespace.com\n"), call("\n"),
+                            call("## Creation Information\n"), call("Creator: Person: actorName (some@mail.com)\n"),
+                            call("Created: 2022-03-10T00:00:00Z\n")])])
+def test_creation_info_writer(creation_info, expected_calls):
+    m = mock_open()
+    with patch("{}.open".format(__name__), m, create=True):
+        with open("foo", "w") as h:
+            write_creation_info(creation_info, h)
```
armintaenzertng, line 40: I get that they are throw-away variables, but can we give `m` and `h` slightly more descriptive names?

tests/spdx/writer/tagvalue/test_creation_info_writer.py
```diff
@@ -0,0 +1,45 @@
...
+
+    m.assert_called_once_with("foo", "w")
+    handle = m()
+    handle.write.assert_has_calls(
+        expected_calls)
```
armintaenzertng, line 45: Unnecessary line break

tests/spdx/writer/tagvalue/test_creation_info_writer.py
```diff
@@ -0,0 +1,45 @@
...
+                                        name="Test document", document_namespace="https://namespace.com",
+                                        created=datetime.datetime(2022, 3, 10)),
+                           [call("SPDXVersion: SPDX-2.3\n"), call("DataLicense: CC0-1.0\n"),
+                            call("SPDXID: SPDXRef-DOCUMENT\n"), call("DocumentName: Test document\n"),
+                            call("DocumentNamespace: https://namespace.com\n"), call("\n"),
+                            call("## Creation Information\n"), call("Creator: Person: actorName (some@mail.com)\n"),
+                            call("Created: 2022-03-10T00:00:00Z\n")])])
+def test_creation_info_writer(creation_info, expected_calls):
+    m = mock_open()
+    with patch("{}.open".format(__name__), m, create=True):
```
armintaenzertng, line 38: Why not use an f-string here?

armintaenzertng (commented):
Thanks for the fix, I have just some stylistic remarks.

tests/spdx/writer/tagvalue/test_creation_info_writer.py
meretp: fixed

tests/spdx/writer/tagvalue/test_creation_info_writer.py
meretp: fixed, I just copied the code from `test_package_writer.py`. I would suggest to change the formatted string there when fixing #394 is fixed, okay?

armintaenzertng: I'd prefer fixing it now so that next time it is not copied from there again ;)

meretp: done

armintaenzertng (approved):

Issue #496, opened 2023-02-28
Release 0.7.1


armintaenzertng:
There have been a few fixes since 0.7.0 which we want to release as 0.7.1.
Things to do beforehand:

- [x] merge #475 and #481
- [x] check for breaking changes
- [x] test the new code
- [x] beta pre-release
- [x] update changelog
- [x] announce new release

ssbarnea:
Apparently there is a a deprecation warning from today's release of setuptools which requires a patch on this library. I will try to make a PR to fix it.
```
DeprecationWarning: pkg_resources is deprecated as an API
```

armintaenzertng:
0.7.1 is up on PyPI

PR #522, opened 2023-03-14
update changelog for v0.7.1

armintaenzertng:
part of #496

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,35 @@
 # Changelog
 
+## v0.7.1 (2023-03-14)
+
+### New features and changes
+
+* added GitHub Actions workflow
+* added requirements.txt
+* added uritools for URI validation
+* Python >= 3.7 is now required
+* json/yaml/xml: added support for empty arrays for hasFiles and licenseInfoFromFiles
+* rdf: fixed writing of multiple packages
+* tag-value: enhanced parsing of snippet ranges to not mix it up with package version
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,34 @@
...
+
+This release was made possible by the following contributors. Thank you very much!
+
+* Christian Decker @chrisdecker1201
+* Marc-Etienne Vargenau @vargenau
+* John Vandenberg @jayvdb
+* Nicolaus Weidner @nicoweidner
+* Meret Behrens @meretp
+* Armin Tänzer @armintaenzertng
+
```
meretp, line 31: The diff between main and the last release also lists maxhbr as one contributor as he added more maintainers to the `pyproject.toml`. So I think he should be listed here as well. But I don't think that this needs to be part of the list of new features and changes

CHANGELOG.md
```diff
@@ -1,5 +1,34 @@
 # Changelog
 
+## v0.7.1 (2023-03-14)
+
+### New features and changes
```
meretp, line 5: As this is only a bug fix release I would expect something like `fixes` here but this heading aligns with the old one and I am not familiar with recommended headings here, so we can leave it as is.

meretp (changes requested):
Thanks for the update! I have two comments, one is only a comment-no change needed and one addition for the `Contributors` list.

meretp (approved):
Thanks for updating the contributors list!

PR #523, opened 2023-03-15
[issue-427] adapt bump methods to allow bumping of actors within other elements

armintaenzertng:
also change `spdx_id_map` to `payload` and fix a bug where `bump_annotation` changed the `creation_info`

### DIFF
--- a/src/spdx3/bump_from_spdx2/annotation.py
+++ b/src/spdx3/bump_from_spdx2/annotation.py
@@ -8,6 +8,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from copy import deepcopy
 
 from spdx.model.annotation import Annotation as Spdx2_Annotation
 from spdx3.bump_from_spdx2.message import print_missing_conversion
@@ -19,6 +20,7 @@ from spdx3.payload import Payload
 def bump_annotation(spdx2_annotation: Spdx2_Annotation, payload: Payload, creation_info: CreationInformation,
                     counter: int):
     spdx_id: str = f"SPDXRef-Annotation-{counter}"
+    creation_info = deepcopy(creation_info)
     creation_info.created = spdx2_annotation.annotation_date
     # creation_info.created_by = bump_actor(spdx2_annotation.annotator)   waiting for entity implementation
     print_missing_conversion("annotation.annotator", 1, "of Entity")
--- a/src/spdx3/bump_from_spdx2/annotation.py
+++ b/src/spdx3/bump_from_spdx2/annotation.py
@@ -13,9 +13,11 @@ from spdx.model.annotation import Annotation as Spdx2_Annotation
 from spdx3.bump_from_spdx2.message import print_missing_conversion
 from spdx3.model.annotation import Annotation, AnnotationType
 from spdx3.model.creation_information import CreationInformation
+from spdx3.payload import Payload
 
 
-def bump_annotation(spdx2_annotation: Spdx2_Annotation, creation_info: CreationInformation, counter: int) -> Annotation:
+def bump_annotation(spdx2_annotation: Spdx2_Annotation, payload: Payload, creation_info: CreationInformation,
+                    counter: int):
     spdx_id: str = f"SPDXRef-Annotation-{counter}"
     creation_info.created = spdx2_annotation.annotation_date
     # creation_info.created_by = bump_actor(spdx2_annotation.annotator)   waiting for entity implementation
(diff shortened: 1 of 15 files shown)


src/spdx3/bump_from_spdx2/annotation.py
```diff
@@ -8,20 +8,24 @@
...
 
 from spdx.model.annotation import Annotation as Spdx2_Annotation
 from spdx3.bump_from_spdx2.message import print_missing_conversion
 from spdx3.model.annotation import Annotation, AnnotationType
 from spdx3.model.creation_information import CreationInformation
+from spdx3.payload import Payload
 
 
-def bump_annotation(spdx2_annotation: Spdx2_Annotation, creation_info: CreationInformation, counter: int) -> Annotation:
+def bump_annotation(spdx2_annotation: Spdx2_Annotation, payload: Payload, creation_info: CreationInformation,
```
meretp, line 20: nit: we incosistently use `creation_info` and `creation_information`. I remember that when I started to implement this I wanted to be consistent with the names for the SPDX3.0 repo and named the class `CreationInformation`, IIRC this is also discussed now, so maybe we can wait with that change and make it consistent once it is decided.

src/spdx3/bump_from_spdx2/annotation.py
```diff
@@ -24,4 +26,4 @@ def bump_annotation(spdx2_annotation: Spdx2_Annotation, creation_info: CreationI
     subject: str = spdx2_annotation.spdx_id
     statement: str = spdx2_annotation.annotation_comment
 
-    return Annotation(spdx_id, creation_info, annotation_type, subject, statement=statement)
+    payload.add_element(Annotation(spdx_id, creation_info, annotation_type, subject, statement=statement))
```
meretp, line 31: Just for the record: Inititally this implementation should have been "target-based", so the return value should be the element to create. With the upcoming Agent  we could still use this approach but with the problem of inlining the Agent Element and problems due to circular import we would need to return the Agent additionally. I think its okay to simply add the bumped element here and make all these bump methods dependant of the payload, so we will always need a `Payload` object to bump any other element from SPDX2.x. As I think that this seems to be the bigger picture for SPDX3.0, I am fine with thesse changes.

tests/spdx3/bump/test_spdx_document_bump.py
```diff
@@ -13,18 +13,18 @@
...
     spdx2_document: Spdx2_Document = document_fixture()
 
-    spdx_id_map: SpdxIdMap = bump_spdx_document(spdx2_document)
+    payload: Payload = bump_spdx_document(spdx2_document)
 
-    write_spdx_id_map(spdx_id_map, sys.stdout)
+    write_payload(payload, sys.stdout)
 
-    assert "SPDXRef-Package" in list(spdx_id_map.get_full_map().keys())
+    assert "SPDXRef-Package" in list(payload.get_full_map().keys())
```
meretp, line 28: You could also just write `"SPDXRef-Package" in payload.get_full_map()` . Probably a matter of taste, which ones better to read.

tests/spdx3/bump/test_spdx_document_bump.py
```diff
@@ -13,18 +13,18 @@
...
     assert len(
-        spdx_id_map.get_full_map().values()) == 6
+        payload.get_full_map().values()) == 6
```
meretp, line 30: No line break needed here. 
And `len(payload.get_full_map()` would also work, as the one above: you can change it, don't have to.

tests/spdx3/bump/test_spdx_document_bump.py
```diff
@@ -13,18 +13,23 @@
...
+
+
+    #this is more of a temporary test to make sure the dates don't get messed up again
```
meretp, line 33: Why is this temporary?

meretp (approved):
Good catch with the bug in `bump_annotation`! Only a few comments from my side which aren't merge-blocking.

tests/spdx3/bump/test_spdx_document_bump.py
armintaenzertng: I imagine we will do more refined testing once the dust has settled.

tests/spdx3/bump/test_spdx_document_bump.py
armintaenzertng: I agree. This was just auto-refactoring here, though... ;)

src/spdx3/bump_from_spdx2/annotation.py
armintaenzertng: Yeah, I stumbled over this, too. Let's wait for the outcome of the renaming discussion.

tests/spdx3/bump/test_spdx_document_bump.py
meretp: Ah yes :+1:

PR #530, opened 2023-03-21
Change error messages related to ISO 8601

### LINKED ISSUE #529: Fix Error Message Related to ISO 8601 format for `creationInfo` `created` field
When a user provides a SPDX document with a time created field that has a fractional second component of an ISO 8601 datetime, `tools-python` returns an error message similar to this:

`Invalid created value '2023-03-17T19:44:10.259Z' must be date in ISO 8601 format.`

After discussion in issue #527 and https://github.com/spdx/ntia-conformance-checker/issues/95, I believe that `tools-python` does not in fact strictly follow ISO 8601, which allows fractional seconds, and instead follows the [SPDX spec](https://spdx.github.io/spdx-spec/v2.3/document-creation-information/#69-created-field), which does not allow fractional seconds.

To help users follow the format expected by the SPDX spec, @zvr suggested this error message instead:

`Invalid created value '2023-03-17T19:44:10.259Z' is not in "YYYY-MM-DDThh:mm:ssZ" format.`

I'm glad to put in a PR if the maintainers are amenable to this change.

jspeed-meyers:
Fix #529.

Because the SPDX specification uses a subset of ISO-8601 for datetime values, error messages that request the user use ISO-8601 dates are incorrect. These changes make clear to the user the exact format, as dictated by the SPDX specification, that dates should be specified in.

I also changed one comment in `utils.py` to make clear to future readers of the codebase what `DATE_ISO_REGEX` represents.

### DIFF
--- a/spdx/parsers/rdf.py
+++ b/spdx/parsers/rdf.py
@@ -36,7 +36,7 @@ ERROR_MESSAGES = {
     "DOC_NAMESPACE_VALUE": 'Invalid DocumentNamespace value {0}, must contain a scheme (e.g. "https:") '
     'and should not contain the "#" delimiter.',
     "LL_VALUE": "Invalid licenseListVersion '{0}' must be of the format N.N where N is a number",
-    "CREATED_VALUE": "Invalid created value '{0}' must be date in ISO 8601 format.",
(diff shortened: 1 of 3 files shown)


armintaenzertng (approved):
Thanks, looks good! :)

PR #531, opened 2023-03-23
Added convenience function to extract SPDX elements by their Id

HarshvMahawar:


src/spdx/document_utils.py
```diff
@@ -9,12 +9,23 @@
...
+from spdx.model.package import Package
+from spdx.model.file import File
 
 
 def get_contained_spdx_element_ids(document: Document) -> List[str]:
     element_ids = [file.spdx_id for file in document.files]
     element_ids.extend([package.spdx_id for package in document.packages])
     element_ids.extend([snippet.spdx_id for snippet in document.snippets])
     return element_ids
+def get_element_from_spdx_id(document: Document, spdx_id: str) -> Union[Package, File, Snippet, None]:
```
armintaenzertng, line 24: PEP8 requires two blank lines between methods.
Which IDE are you using for your python development? There might be a plugin for your IDE to detect PEP8 violations.

armintaenzertng (commented):
Thanks for your contribution! :)
If you like, you can also include a unit test for this under `tests/spdx/test_document_utils.py`. If not, I'm happy to include it after this PR is merged.

src/spdx/document_utils.py
HarshvMahawar: I am using VSCode

armintaenzertng (approved):
Thanks, looks good! :)

PR #534, opened 2023-03-24
small changes to tag-value lexer/parser

armintaenzertng:
change `DATE` to `ISO8601_DATE`, don't inherit from `object`

### DIFF
--- a/src/spdx/parser/tagvalue/lexer.py
+++ b/src/spdx/parser/tagvalue/lexer.py
@@ -14,7 +14,7 @@ from ply import lex
 from ply.lex import TOKEN
 
 
-class SPDXLexer(object):
+class SPDXLexer:
     reserved = {
         # Top level fields
         "SPDXVersion": "DOC_VERSION",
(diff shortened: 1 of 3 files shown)


src/spdx/parser/tagvalue/lexer.py
```diff
@@ -107,7 +107,7 @@ class SPDXLexer(object):
                  "UNKNOWN_TAG",
                  "ORGANIZATION_VALUE",
                  "PERSON_VALUE",
-                 "DATE",
+                 "ISO8601_DATE",
```
meretp, line 110: As there was recently a [discussion](https://github.com/spdx/ntia-conformance-checker/issues/95) about the date not being exatly ISO8601 I am wondering if we should really name it like that.

zvr: Let's be clear: it **is** an ISO-8601 formatted date.

It doesn't allow _all_ variants of ISO-8601, which include `20230328T122341.123` or `20230328T12XXXX` or `2023-33`. You would find very few implementations handling all these variants.

meretp (approved):
lgtm

PR #537, opened 2023-03-29
Add python 3.11 to GitHub Action

meretp:
As we support Python 3.11 we should also test against it in our GitHub Action.
Additionally I added a push to the branch as a trigger for this workflow. When merging this we will need to adapt this.


armintaenzertng (approved):
Thanks :)

PR #541, opened 2023-03-29
Merged the refactoring branch into main

maxhbr:
This is **not to be merged**, but instead its HEAD 3f2619c3da8d62aa1966487ac00a4d0512b5a0e7 is the relevant merge commit should directly be pushed to main.

@goneall : can you open the branch protection rules for a short window so that I can push or otherwise you can push it (if admins can ignore branch protection rules in this repo)?


goneall (approved):
I didn't review all 365 files, but the structure looks good and I agree we should move the refactored branch into main.

PR #543, opened 2023-03-29
[issue-542] use the documents namespace as prefix

### LINKED ISSUE #542: Document node in RDF output has no unique ID
The RDF writer produces a `SpdxDocument` instance that always has the identifier `http://www.spdx.org/tools#SPDXRef-DOCUMENT`. When merging multiple SPDX documents from across multiple distributions, this means that e.g. all package dependencies, the names, the version, the license, creationInfo etc. will be attached to a single node in the RDF graph.

https://github.com/spdx/tools-python/blob/73f3f462591894cac8ef11e01ebec7e702e33c25/spdx/writers/rdf.py#L987

It makes much more sense to use the `name` of the document to construct an identifier as in this line:

https://github.com/spdx/tools-python/blob/73f3f462591894cac8ef11e01ebec7e702e33c25/spdx/writers/rdf.py#L996

meretp:
Note that this fix won't be merged to main as we are in the process of merging the refactored version to main. This PR will instead be merged to the v.0.7.1 branch. 

fixes #542

### DIFF
--- a/tests/data/doc_write/rdf-mini.json
+++ b/tests/data/doc_write/rdf-mini.json
@@ -3,7 +3,7 @@
     "@xmlns:ns1": "http://spdx.org/rdf/terms#",
     "@xmlns:rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
     "ns1:SpdxDocument": {
-      "@rdf:about": "http://www.spdx.org/tools#SPDXRef-DOCUMENT",
+      "@rdf:about": "None#SPDXRef-DOCUMENT",
       "ns1:specVersion": "SPDX-2.1",
       "ns1:dataLicense": {
         "@rdf:resource": "http://spdx.org/licenses/CC0-1.0"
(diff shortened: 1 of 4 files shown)


tests/data/doc_write/rdf-mini.json
```diff
@@ -3,7 +3,7 @@
     "@xmlns:ns1": "http://spdx.org/rdf/terms#",
     "@xmlns:rdf": "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
     "ns1:SpdxDocument": {
-      "@rdf:about": "http://www.spdx.org/tools#SPDXRef-DOCUMENT",
+      "@rdf:about": "None#SPDXRef-DOCUMENT",
```
armintaenzertng, line 6: `None` is not a valid namespace, I believe... Why not use the one that the rest of the test files use?

meretp: This test is based on the function `_get_mini_doc()` in test_document.py. This minimal document seems to be from some earlier SPDX specification where the namespace probably wasn't mandatory. I can add the namespace to the function as a hoepfully minimal-invasive fix..

armintaenzertng: Nah, if that is some special, long forgotten case, I don't think we should bother.

armintaenzertng (approved):

PR #544, opened 2023-03-30
[issue-532] Add linter action

meretp:
This PR adds a GitHub Action to check the code style with `black`, `isort` and `flake8`. The new GitHub Action is expected to fail as the current code doesn't comply with these linters. A second PR will follow to fix this.


armintaenzertng (approved):
Thanks, I just tried it and it worked mostly except for very long strings. Is this something that has to be done manually?

meretp:
Unfortunately yes. With a quick research I found that there are some tools that could do this automatically but as this would probably need more work (setting up the tool and so on), I decided to do it manually. I would assume that this is a justifiable effort in the future.

armintaenzertng:
A fix of this is part of #546

PR #549, opened 2023-03-30
[issue-547] fix rdf parser

### LINKED ISSUE #547: RDF parser omits ID of extracted licensing info
Parsing an RDF document like the `SPDXRdfExample-v2.2.spdx.rdf.xml` results in extracted licensing infos such as this:
```
ExtractedLicensingInfo(license_id=None, extracted_text='"THE BEER-WARE LICENSE" (Revision 42):\nphk@FreeBSD.ORG wrote this file. As long as you retain this notice you\ncan do whatever you want with this stuff. If we meet some day, and you think this stuff is worth it, you can buy me a beer in return Poul-Henning Kamp', license_name='Beer-Ware License (Version 42)', cross_references=['http://people.freebsd.org/~phk/'], comment='The beerware license has a couple of other standard variants.')
```
That is, `license_id=None` while it should be `license_id="LicenseRef-Beerware-4.2"`

meretp:
Use licenseID from extracted licensing info URIRef if no licenseId node is provided.

fixes #547

### DIFF
--- a/tests/spdx/parser/rdf/test_extracted_licensing_info_parser.py
+++ b/tests/spdx/parser/rdf/test_extracted_licensing_info_parser.py
@@ -3,21 +3,51 @@
 # SPDX-License-Identifier: Apache-2.0
 import os
 
+import pytest
 from rdflib import RDF, Graph
 
 from spdx.parser.rdf.extracted_licensing_info_parser import parse_extracted_licensing_info
 from spdx.rdfschema.namespace import SPDX_NAMESPACE
 
 
-def test_parse_extracted_licensing_info():
+@pytest.mark.parametrize(
+    "license_id, extracted_text, comment, license_name, cross_references",
+    [
+        (
+            "LicenseRef-1",
+            "extractedText",
+            "licenseComment",
+            "licenseName",
+            ["https://see.also"],
+        ),
+        (
+            "LicenseRef-2",
+            "extractedText",
+            "licenseComment",
+            "another license",
+            ["https://see.also"],
+        ),
+    ],
+)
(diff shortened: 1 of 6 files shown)


tests/spdx/parser/rdf/test_extracted_licensing_info_parser.py
armintaenzertng, line 1: We should have two tests, one for each case of how the license Id could be present.

tests/spdx/parser/rdf/test_extracted_licensing_info_parser.py
```diff
@@ -10,21 +10,51 @@
...
+            "LicenseRef-2",
+            "extractedText",
+            "licenseComment",
+            "another license",
+            ["https://see.also"],
+        ),
+    ],
+)
+def test_parse_extracted_licensing_info(
+    parent_node_object, license_id, extracted_text, comment, license_name, cross_references
```
armintaenzertng, line 42: `parent_node_object` seems to be unused

tests/spdx/parser/rdf/test_extracted_licensing_info_parser.py
```diff
@@ -10,21 +10,51 @@
...
+            "licenseName",
+            ["https://see.also"],
+        ),
+        (
+            SPDX_NAMESPACE.Package,
+            "LicenseRef-2",
+            "extractedText",
+            "licenseComment",
+            "another license",
+            ["https://see.also"],
```
armintaenzertng, line 37: It might be helpful to give a quick explanation of the issue at hand for future developers, i.e. what's the difference between the two test cases.

meretp: Done

armintaenzertng (approved):

PR #552, opened 2023-03-31
shorten line lengths in README

armintaenzertng:
While reading through the README on our repo's main page, I noticed that you have to scroll to read the complete inline comments in the Python example. This should not be an issue anymore, now.


meretp (approved):
Thanks!

PR #554, opened 2023-04-03
[spdx-3.0] add package_version

armintaenzertng:
As defined in https://github.com/spdx/spdx-3-model/blob/main/model/Software/Classes/Package.md.
During bumping, this will be filled with the content of `spdx2_package.version`.

### DIFF
--- a/src/spdx3/bump_from_spdx2/package.py
+++ b/src/spdx3/bump_from_spdx2/package.py
@@ -18,8 +18,7 @@ def bump_package(
     spdx_id = "#".join([document_namespace, spdx2_package.spdx_id])
     name = spdx2_package.name
     download_location = handle_no_assertion_or_none(spdx2_package.download_location, "package.download_location")
-    # package2.version -> ?
-    print_missing_conversion("package2.version", 0)
+    package_version = spdx2_package.version
     # package.file_name -> ?
     print_missing_conversion("package2.file_name", 0)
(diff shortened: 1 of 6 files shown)


src/spdx3/bump_from_spdx2/package.py
```diff
@@ -54,13 +53,14 @@ def bump_package(
...
-            homepage=homepage,
             summary=summary,
             description=description,
             comment=comment,
+            verified_using=integrity_methods,
             originated_by=originated_by_spdx_id,
             package_purpose=package_purpose,
+            package_version=package_version,
+            download_location=download_location,
+            homepage=homepage,
```
armintaenzertng, line 64: I'm not sure why the first black sorting (in the PR just a few minutes ago) used a different sorting than the one I did now in this PR

meretp: I also don't know why this happens. The only thing I could imagine: By adding `package_version` black was triggered in some different way and aligned the order of arguments according to the definition of `Package.__init__()`. But I also don't know why this wasn't triggered in the first PR.

meretp (changes requested):
Thanks for adding this, I don't have any change requests regarding your changes but could you please also update the `README.md` and the debug writer.

armintaenzertng:
> could you please also update the `README.md` and the debug writer.
Done

PR #559, opened 2023-04-03
[SPDX-3.0] add creation_info comment

armintaenzertng:
As seen in https://github.com/spdx/spdx-3-model/blob/main/model/Core/Classes/CreationInformation.md.
`spdx2_creation_info.document_comment` converts into this.


meretp (approved):
Thanks!

PR #560, opened 2023-04-03
Fix license identifiers in SPDX-3 prototype branch

armintaenzertng:
Based on this comment: https://github.com/spdx/tools-python/pull/557#discussion_r1155784683


meretp (approved):
Thanks! I don't know if there is a semantically equivalent to this in English, but my "inner Monk" is very happy right now! :nerd_face:

PR #562, opened 2023-04-03
[issue-558] add visualization feature

meretp:
This PR adds some functionality to visualize the structure of a SPDX document. The visualization includes the SPDXIDs of all elements in the document (`Package`, `File`, `Snippet` and any external SPDXIDs) and  their connection based on the provided relationships. The new logic converts a `Document` to a `pygraphviz` `AGraph` which can either be stored in a `.dot` file or rendered to an image e.g. a `.png`.


src/common/visualization/visualization.py
armintaenzertng, line 1: I don' think this belongs to `src/common` as it would not work with an SPDX3 document

src/common/visualization/visualization.py
```diff
@@ -0,0 +1,81 @@
...
+from spdx.document_utils import get_contained_spdx_element_ids, get_element_from_spdx_id
+from spdx.model.document import Document
+from spdx.model.relationship import Relationship
+
+
+def visualize_document(document: Document, visualize: str) -> None:
+    try:
+        from networkx.drawing import nx_agraph
+    except ImportError:
+        raise ImportError
```
armintaenzertng, line 19: This doesn't seem to make sense (see also below)

src/common/visualization/visualization.py
```diff
@@ -0,0 +1,81 @@
...
+    for relationship in relationships:
+        relationships_by_spdx_id.setdefault(relationship.spdx_element_id, []).append(relationship)
+
+    return relationships_by_spdx_id
+
+
+def _color_nodes(graph: DiGraph) -> None:
+    for node in graph.nodes():
+        if "_" in node:
+            graph.add_node(node, color="lightgreen", label=node.split("_", 1)[-1])
```
armintaenzertng, line 77: I'd like a comment here to provide information about which cases this "_"-case refers to

README.md
```diff
@@ -78,6 +80,15 @@ instead of `bin`.
...
 * For help use `pyspdxtools --help`
 
+3. **VISUALIZATION** (optional feature)
+
+* Make sure you install the optional dependencies `networkx` and `pygraphviz`. To do so run `pip install ".[visualization]"`.
+* Use `pyspdxtools -i <input_file> --visualize <output_file>` where `<output_file>` is a valid output for `pygraphviz` (check 
+  the documentation [here](https://pygraphviz.github.io/documentation/stable/reference/agraph.html#pygraphviz.AGraph.draw)). 
+* If you are using a source distribution, try running
+  `pyspdxtools -i tests/data/formats/SPDXJSONExample-v2.3.spdx.json --visualize SPDXJSONExample-v2.3.spdx.png` to generate 
+  a png with an overview of the structure of the example file.
```
armintaenzertng, line 90: We should provide an example of this somewhere so users will know if this feature is something they want before they actually have to install all this.

src/common/visualization/visualization.py
```diff
@@ -0,0 +1,81 @@
...
+try:
+    from networkx import DiGraph
+except ImportError:
+    DiGraph = None
+from spdx.document_utils import get_contained_spdx_element_ids, get_element_from_spdx_id
+from spdx.model.document import Document
+from spdx.model.relationship import Relationship
+
+
+def visualize_document(document: Document, visualize: str) -> None:
```
armintaenzertng, line 15: Maybe it's just because I haven't used graph libraries in python before, but I find it rather unclear what kind of strings `visualize` would take or what it actually does. (Sounds more like a Boolean to me.)

src/common/visualization/visualization.py
```diff
@@ -0,0 +1,81 @@
...
+
+    return graph
+
+
+def _get_relationships_by_spdx_id(
+    relationships: List[Relationship],
+) -> Dict[str, List[Relationship]]:
+    relationships_by_spdx_id: Dict[str, List[Relationship]] = dict()
+    for relationship in relationships:
+        relationships_by_spdx_id.setdefault(relationship.spdx_element_id, []).append(relationship)
```
armintaenzertng, line 69: I don' think these two lines warrant their own method.

src/common/visualization/visualization.py
```diff
@@ -0,0 +1,81 @@
...
+    try:
+        from networkx import DiGraph
+    except ImportError:
+        raise ImportError
+    graph = DiGraph()
+    graph.add_node(document.creation_info.spdx_id, element=document.creation_info)
+
+    contained_elements: List[str] = get_contained_spdx_element_ids(document)
+    contained_element_nodes = [
+        (spdx_id, {"element": get_element_from_spdx_id(document, spdx_id)}) for spdx_id in contained_elements
```
armintaenzertng, line 37: Given that you quite often call `get_element_from_spdx_id` in the code below, I think it would be beneficial for performance if we introduced another `document_utils` method which returns a dictionary that contains all SPDX Ids mapped to their corresponding element. That is, it returns something like the payload in SPDX3.

src/common/visualization/visualization.py
```diff
@@ -0,0 +1,81 @@
...
+    ]
+    graph.add_nodes_from(contained_element_nodes)
+
+    relationships_by_spdx_id: Dict[str, List[Relationship]] = _get_relationships_by_spdx_id(document.relationships)
+
+    for spdx_id, relationships in relationships_by_spdx_id.items():
+        if spdx_id not in graph.nodes():
+            graph.add_node(spdx_id, element=get_element_from_spdx_id(document, spdx_id))
```
armintaenzertng, line 45: Didn't you already add all contained elements in line 39? So `get_element_from_spdx_id(document, spdx_id)` always yields `None` here, doesn't it?

src/common/visualization/visualization.py
```diff
@@ -0,0 +1,81 @@
...
+            graph.add_node(relationship_node_key, comment=relationship.comment)
+            graph.add_edge(relationship.spdx_element_id, relationship_node_key)
+            # if the related spdx element is SpdxNone or SpdxNoAssertion we need a
+            # type conversion
+            related_spdx_element_id = str(relationship.related_spdx_element_id)
+
+            if related_spdx_element_id not in graph.nodes():
+                graph.add_node(
+                    related_spdx_element_id,
+                    element=get_element_from_spdx_id(document, spdx_id),
```
armintaenzertng, line 57: Same question as above.

tests/common/test_visualization.py
```diff
@@ -0,0 +1,163 @@
+# SPDX-FileCopyrightText: 2023 TNG Technology Consulting GmbH <https://www.tngtech.com>
```
maxhbr, line 1:
```suggestion
# SPDX-FileCopyrightText: 2023 spdx contributors
```

tests/common/test_visualization.py
```diff
@@ -0,0 +1,163 @@
...
+    nodes_count: int,
+    edges_count: int,
+    relationship_node_keys: List[str],
+) -> None:
+    document = parse_file(str(Path(__file__).resolve().parent.parent / "spdx" / "data" / "formats" / file_name))
+    graph = generate_graph_from_spdx(document)
+
+    assert document.creation_info.spdx_id in graph.nodes()
+
+    assert graph.number_of_nodes() == nodes_count
```
armintaenzertng, line 70: unnecessary blank line

tests/common/test_visualization.py
```diff
@@ -0,0 +1,163 @@
...
+    )
+
+
+def test_complete_unconnected_graph() -> None:
+    document = _create_minimal_document()
+    document.packages += [
+        Package(
+            spdx_id="SPDXRef-Package-C",
+            name="Package without connection to document",
+            download_location="https://download.location.com",
```
armintaenzertng, line 114: Why do you provide a download location here?

armintaenzertng (changes requested):
Thanks for this! :)
I've left a few change requests and comments below.

zvr:
Ehmmm... a more general question: why did you choose to use `pygraphviz` and `networkx` instead of the much simpler `graphviz` module? 

I mean, it's nice to have everything inside your code, but you then will have to eventually expose all functions somehow (e.g., produce SVG, produce PDF, scale factor, use different layout engine, etc., etc.)

The alternative would be:
```Python
import graphviz
g = graphviz.Digraph('build_recorder RDF data', filename=fname_out) 
for e in element_list:
    g.node(e.id, label=e.name)
for r in relationship_list:
    g.edge(r.from, r.to)
g.save()
# g.view() if you want
```

maxhbr:
We also want to expose functions to convert to networkx. This is helpful to work with the complex data.

zvr:
Ah, ok. Then this is not about "visualization", the requirement is to have the graph in networkx.

src/common/visualization/visualization.py
meretp: This is the file where the generated graph - or rendered picture depending on the file extension will be written to. I will rename this variable.

src/common/visualization/visualization.py
meretp: Yes, this line will add any external spdx_id if present so `get_element_from_spdx_id` yields `None`. For the later use of the graph, it would be helpful if the node has some kind of flag indicating that the node represents an element (including external elements), so I would keep the node attribute `element` and fill it with `None`.

tests/common/test_visualization.py
meretp: Basically to make this package valid, which was needed in an earlier development. I think I will just use a `package_fixture` here.

tests/common/test_visualization.py
meretp: fixed

meretp:
According to the comments I restructured and renamed the methods and the description in the markdown. This PR now adds  functionality to create a relationship graph from the document which can be rendered to a picture for visualization but doesn't have to be. I hope this fits the comments.
I have one open question: 
How to include this feature in the cli? The current implementation uses the argument `--graph_generation` and expects a file path as input where the generated graph - or rendered picture- will be stored to. An alternative approach would be to use `--graph_generation` only as a flag and use `--outfile` for the file path which would be on the one hand more consistent, on the other hand could also be confusing as `--outfile` is reserved for the conversion of documents. What do you think?

src/common/visualization/visualization.py
meretp: fixed

README.md
meretp: I added an example png in the readme, please have a look if this fits your intention.

src/common/visualization/visualization.py
meretp: added

README.md
```diff
@@ -78,6 +80,18 @@ instead of `bin`.
   
 * For help use `pyspdxtools --help`
 
+3. **GRAPH GENERATION** (optional feature)
+
+* This feature generates a graph representing all elements in the SPDX document and their connections based on the provided
+  relationships. The graph can be rendered to a picture. Below is an example for the file `tests/data/formats/SPDXJSONExample-v2.3.spdx.json`:
+![SPDXJSONExample-v2.2.spdx.png](SPDXJSONExample-v2.2.spdx.png)
```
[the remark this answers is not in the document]
meretp: Yes, of course, I added the file and updated the README

armintaenzertng (commented):
Just a few more comments regarding wording.

armintaenzertng:
> How to include this feature in the cli? The current implementation uses the argument --graph_generation and expects a file path as input where the generated graph - or rendered picture- will be stored to. An alternative approach would be to use --graph_generation only as a flag and use --outfile for the file path which would be on the one hand more consistent, on the other hand could also be confusing as --outfile is reserved for the conversion of documents. What do you think?

One "advantage" of the current approach is that you can do both conversion and graph generation in one. I'm not sure, though, if this is actually something to be avoided. I don't have strong feelings either way.

src/spdx/graph_generation.py
```diff
@@ -38,42 +33,40 @@ def generate_graph_from_spdx(document: Document) -> DiGraph:
     ]
     graph.add_nodes_from(contained_element_nodes)
 
-    relationships_by_spdx_id: Dict[str, List[Relationship]] = _get_relationships_by_spdx_id(document.relationships)
+    relationships_by_spdx_id: Dict[str, List[Relationship]] = dict()
+    for relationship1 in document.relationships:
```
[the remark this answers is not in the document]
meretp: just a typo :smile:

src/spdx/graph_generation.py
```diff
@@ -0,0 +1,76 @@
...
+        for relationship in relationships:
+            relationship_node_key = relationship.spdx_element_id + "_" + relationship.relationship_type.name
+            graph.add_node(relationship_node_key, comment=relationship.comment)
+            graph.add_edge(relationship.spdx_element_id, relationship_node_key)
+            # if the related spdx element is SpdxNone or SpdxNoAssertion we need a type conversion
+            related_spdx_element_id = str(relationship.related_spdx_element_id)
+
+            if related_spdx_element_id not in graph.nodes():
+                # this will add any external spdx_id to the graph where we have no further information about
+                # the element to indicate that this node represents an element we add the attribute "element"
```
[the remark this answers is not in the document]
meretp: done

src/spdx/graph_generation.py
```diff
@@ -38,42 +33,40 @@ def generate_graph_from_spdx(document: Document) -> DiGraph:
...
-        relationships_by_spdx_id.setdefault(relationship.spdx_element_id, []).append(relationship)
-
-    return relationships_by_spdx_id
-
-
 def _color_nodes(graph: DiGraph) -> None:
     for node in graph.nodes():
         if "_" in node:
+            # nodes representing a RelationshipType are concatenated with the spdx_element_id
+            # to only see the RelationshipType when rendering the graph to a picture we add
```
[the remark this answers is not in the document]
meretp: done

README.md
```diff
@@ -80,13 +80,16 @@ instead of `bin`.
...
-3. **VISUALIZATION** (optional feature)
+3. **GRAPH GENERATION** (optional feature)
 
-* Make sure you install the optional dependencies `networkx` and `pygraphviz`. To do so run `pip install ".[visualization]"`.
-* Use `pyspdxtools -i <input_file> --visualize <output_file>` where `<output_file>` is a valid output for `pygraphviz` (check 
+* This feature generates a graph representing all elements in the SPDX document and their connections based on the provided
+  relationships. The graph can be rendered to a picture. Below is an example for the file `tests/data/formats/SPDXJSONExample-v2.3.spdx.json`:
+![SPDXJSONExample-v2.3.spdx.png](assets/SPDXJSONExample-v2.3.spdx.png)
+* Make sure you install the optional dependencies `networkx` and `pygraphviz`. To do so run `pip install ".[graph_generation]"`.
+* Use `pyspdxtools -i <input_file> --graph_generation <output_file>` where `<output_file>` is a valid output for `pygraphviz` (check
```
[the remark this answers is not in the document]
meretp: done

armintaenzertng (approved):
Thanks! :)

zvr:
My vote is for 

> use --graph_generation only as a flag and use --outfile for the file path

It's better to be consistent with treating `outfile` as the output of the process. Yes, it limits that you cannot convert and generate the graph on the same step, but I don't think that would be a common use case.

Addiotnally, please rethink the ` --generate_graph` CLI flag and try to avoid the underscore (isn't `--graph` enough? else, use `-`).

Ehhmmm, wait a minute: is it an option or a subcommand? Apologies for not checking, but is the rest of functionality invoked as `--validate` , `--convert`, etc. or as commands (like `git foo`).

meretp:
> Ehhmmm, wait a minute: is it an option or a subcommand? Apologies for not checking, but is the rest of functionality invoked as --validate , --convert, etc. or as commands (like git foo).

The rest is invoked as options, e.g. `--infile` (`-i` as shortform). According to this I will add an option `--graph` which generates the graph and writes it to the file provided with `--outfile`.

src/spdx/graph_generation.py
```diff
@@ -0,0 +1,76 @@
...
+            # to indicate that this node represents an element we add the attribute "element"
+            graph.add_node(spdx_id, element=None)
+        for relationship in relationships:
+            relationship_node_key = relationship.spdx_element_id + "_" + relationship.relationship_type.name
+            graph.add_node(relationship_node_key, comment=relationship.comment)
+            graph.add_edge(relationship.spdx_element_id, relationship_node_key)
+            # if the related spdx element is SpdxNone or SpdxNoAssertion we need a type conversion
+            related_spdx_element_id = str(relationship.related_spdx_element_id)
+
+            if related_spdx_element_id not in graph.nodes():
```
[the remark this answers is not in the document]
meretp: Your interpretation for the current implementation is correct. I don't have an opinion on how it should be. In the current use case we would transform the generated graph to a tree, so I didn't worry about it. But probably it would be better (clearer, in relation to a picture) to have individual NOASSERTION nodes for each relationship.

src/spdx/clitools/pyspdxtools.py
```diff
@@ -72,7 +79,20 @@ def main(infile: str, outfile: str, version: str, novalidation: bool):
             else:
                 logging.info("The document is valid.")
 
-        if outfile and outfile != "-":
+        if outfile == "-":
+            tagvalue_writer.write_document(document, sys.stdout)
+
+        elif graph:
+            try:
+                export_graph_from_document(document, outfile)
```
armintaenzertng, line 87: Maybe insert a catch that `outfile` must be set, i.e. not `None`, and inform the user about it?

armintaenzertng (dismissed):
Looks good to me now, only a minor remark

src/spdx/clitools/pyspdxtools.py
maxhbr: sounds like a nice to have. Would merge it now

armintaenzertng (approved):

PR #563, opened 2023-04-04
[issue-561] fix rdf parser

### LINKED ISSUE #561: rdf parser doesn't parse hasFile tag
According to the [ontology](https://spdx.org/rdf/terms/#d4e2963) a package can have multiple `hasFile` nodes which should be translated to a `package CONTAINS file` relationship while parsing a rdf file. E.g. the package `SPDXRef-Package` in the provided [2.2 example file](https://github.com/spdx/tools-python/blob/ac9e62fd970ad40aba8fb7fd2988691b35e54b88/tests/spdx/data/formats/SPDXRdfExample-v2.2.spdx.rdf.xml#L2657) has two `hasFiles` childs. This relation is omitted while poarsing the file which is a bug.

meretp:
This PR adds some logic to parse implicit "hasFiles" and "describesPackage" relationships. 

fixes #561

### DIFF
--- a/src/spdx/parser/rdf/graph_parsing_functions.py
+++ b/src/spdx/parser/rdf/graph_parsing_functions.py
@@ -2,12 +2,12 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 from enum import Enum
-from typing import Any, Callable, Optional, Type
+from typing import Any, Callable, Optional, Tuple, Type, Union
 
-from rdflib import Graph, URIRef
+from rdflib import RDF, Graph, URIRef
 from rdflib.exceptions import UniquenessError
 from rdflib.namespace import NamespaceManager
-from rdflib.term import Node
+from rdflib.term import BNode, Literal, Node
 
 from spdx.casing_tools import camel_case_to_snake_case
 from spdx.model.spdx_no_assertion import SPDX_NO_ASSERTION_STRING, SpdxNoAssertion
@@ -102,3 +102,46 @@ def remove_prefix(string: str, prefix: str) -> str:
     if string.startswith(prefix):
         return string[len(prefix) :]
     return string
+
+
+def get_correctly_typed_triples(
+    logger: Logger,
+    graph: Graph,
+    subject: Optional[Node] = None,
+    predicate: Optional[Node] = None,
+    _object: Optional[Node] = None,
+) -> Tuple[Union[BNode, URIRef], Node, Union[BNode, Literal, URIRef]]:
+    # this is a helper method to cast some rdf types from graph.triples() to be compatible with the
+    # code that follows
+    for s, p, o in graph.triples((subject, predicate, _object)):
+        if not isinstance(s, (BNode, URIRef)):
+            logger.append(
+                f"Warning: Subject {s} should be of type BNode or URIRef, but is {type(s).__name__}. "
+                f"This might lead to a failure."
+            )
+        if not isinstance(o, (BNode, Literal, URIRef)):
+            logger.append(
+                f"Warning: Object {o} should be of type BNode, Literal or URIRef, but is {type(o).__name__}. "
+                f"This might lead to a failure."
+            )
+        yield s, p, o
+
+
+def get_value_from_graph(
+    logger: Logger,
+    graph: Graph,
+    subject: Optional[Node] = None,
+    predicate: Optional[Node] = RDF.value,
+    _object: Optional[Node] = None,
+    default: Optional[Any] = None,
+    _any: Optional[bool] = True,
+) -> Optional[Union[URIRef, Literal, BNode]]:
+    # this is a helper method to cast some rdf types from graph.value() to be compatible with the
+    # code that follows
+    value = graph.value(subject=subject, predicate=predicate, object=_object, default=default, any=_any)
+    if value and not isinstance(value, (URIRef, Literal, BNode)):
+        logger.append(
+            f"Warning: Node {value} should be of type BNode, Literal or URIRef, but is {type(value).__name__}. "
+            f"This might lead to a failure."
+        )
+    return value
--- a/src/spdx/parser/rdf/rdf_parser.py
+++ b/src/spdx/parser/rdf/rdf_parser.py
@@ -1,9 +1,12 @@
 # SPDX-FileCopyrightText: 2023 spdx contributors
 #
 # SPDX-License-Identifier: Apache-2.0
+from typing import Any, Dict
+
 from rdflib import RDF, Graph
 
 from spdx.model.document import Document
+from spdx.model.relationship import RelationshipType
 from spdx.parser.error import SPDXParsingError
 from spdx.parser.logger import Logger
 from spdx.parser.parsing_functions import construct_or_raise_parsing_error, raise_parsing_error_if_logger_has_messages
@@ -12,7 +15,7 @@ from spdx.parser.rdf.creation_info_parser import parse_creation_info
 from spdx.parser.rdf.extracted_licensing_info_parser import parse_extracted_licensing_info
 from spdx.parser.rdf.file_parser import parse_file
 from spdx.parser.rdf.package_parser import parse_package
-from spdx.parser.rdf.relationship_parser import parse_relationship
+from spdx.parser.rdf.relationship_parser import parse_implicit_relationship, parse_relationship
 from spdx.parser.rdf.snippet_parser import parse_snippet
 from spdx.rdfschema.namespace import SPDX_NAMESPACE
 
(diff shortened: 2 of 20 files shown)


src/spdx/parser/rdf/rdf_parser.py
```diff
@@ -62,6 +65,21 @@ def translate_graph_to_document(graph: Graph) -> Document:
...
         parsed_fields[element] = elements
 
+    for element, triple, relationship_type in [
+        ("has_files", (None, SPDX_NAMESPACE.hasFile, None), RelationshipType.CONTAINS),
+        ("describes_package", (None, SPDX_NAMESPACE.describesPackage, None), RelationshipType.DESCRIBES),
+    ]:
+        for parent_node, _, element_node in graph.triples(triple):
+            try:
+                relationship = parse_implicit_relationship(
+                    element_node, graph, parent_node, creation_info.document_namespace, relationship_type
```
meretp, line 75: In this line PyCharm complains about a wrong type: "Expected type 'URIRef', got 'Node' instead". I don't know how to solve this, the return type of `graph.triples(..)` is `Tuple[Node, Node, Node]`,  `parse_implicit_relationship` expects a `URIRef`. Any suggestions how to fix are much appreciated.

src/spdx/parser/rdf/relationship_parser.py
```diff
@@ -49,3 +49,25 @@ def parse_relationship(
...
 
     return relationship
+
+
+def parse_implicit_relationship(
+    related_spdx_element_node: URIRef,
+    graph: Graph,
+    parent_node: URIRef,
+    doc_namespace: str,
+    relationship_type: RelationshipType,
```
armintaenzertng, line 59: I find the argument order here confusing, as the "normal" order would be 
`parent_node`, `relationship_type`, `related_spdx_element_node`.
I also took some time to understand that `parent_node` refers to the SPDX element on the left side of the relationship, but I'm currently at a loss for a better name...

src/spdx/parser/rdf/rdf_parser.py
```diff
@@ -62,6 +65,21 @@ def translate_graph_to_document(graph: Graph) -> Document:
...
+                )
+                if relationship not in parsed_fields["relationships"]:
```
armintaenzertng, line 77: `parsed_fields["relationships"]` could be `None`, so we should use the default method here.

src/spdx/parser/rdf/rdf_parser.py
armintaenzertng: I think your proposal is the right way to go. We can't expect a "random" triple from the parsed RDF graph to be a `URIRef`, so we have to make sure that we get the correct type. I'm curious, though, why this doesn't pop up in other uses of `graph.triples()`.

armintaenzertng (changes requested):
Thanks for spotting and fixing this! :)
I have a few small remarks.

src/spdx/parser/rdf/rdf_parser.py
meretp: `parsed_fields["relationships"]` can't be `None`. If there are no relationships in the document it will be an empty list. See Line 6o and 66.

src/spdx/parser/rdf/relationship_parser.py
meretp: Alright, I will change the order. In the first place, I thought I use the same for-loop as for the relationships in `rdf_parser.py` which would have needed this order but as this isn't possible, I can rearrange this.

src/spdx/parser/rdf/rdf_parser.py
meretp: I added a wrapper method for the `graph.triples(..)` and also for `graph.value(..)` to ensure types at some places. I decided to write a Warning in the logger if the type doesn't match as the code would probably run either way.

src/spdx/parser/rdf/rdf_parser.py
```diff
@@ -62,6 +65,21 @@ def translate_graph_to_document(graph: Graph) -> Document:
                 logger.extend(err.get_messages())
         parsed_fields[element] = elements
 
+    for element, triple, relationship_type in [
```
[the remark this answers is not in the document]
meretp: deleted

src/spdx/parser/rdf/relationship_parser.py
```diff
@@ -49,3 +49,25 @@ def parse_relationship(
...
+) -> Relationship:
+    logger = Logger()
+    spdx_element_id = parse_spdx_id(parent_node, doc_namespace, graph)
+    related_spdx_element = parse_spdx_id(related_spdx_element_node, doc_namespace, graph)
+    raise_parsing_error_if_logger_has_messages(logger, "Relationship")
```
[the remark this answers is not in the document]
meretp: Correct, a typical copy-paste-don't care- code :see_no_evil:

src/spdx/parser/rdf/graph_parsing_functions.py
```diff
@@ -102,3 +102,44 @@ def remove_prefix(string: str, prefix: str) -> str:
...
+
+
+def get_correctly_typed_triples(
+    logger: Logger,
+    graph: Graph,
+    subject: Optional[Node] = None,
+    predicate: Optional[Node] = None,
+    _object: Optional[Node] = None,
+) -> Tuple[URIRef, Node, Union[BNode, Literal, URIRef]]:
+    # this is a helper method to enforce some rdf types for the triples to bo compatible with the
```
armintaenzertng, line 114:
```suggestion
    # this is a helper method to cast some rdf types from graph.triples() to be compatible with the
```

src/spdx/parser/rdf/graph_parsing_functions.py
```diff
@@ -102,3 +102,44 @@ def remove_prefix(string: str, prefix: str) -> str:
...
+    # code that follows
+    for s, p, o in graph.triples((subject, predicate, _object)):
+        if not isinstance(s, (BNode, URIRef)):
```
armintaenzertng, line 117: We allow `BNode` here but the type hint for the subject is only `URIRef`

src/spdx/parser/rdf/graph_parsing_functions.py
```diff
@@ -102,3 +102,44 @@ def remove_prefix(string: str, prefix: str) -> str:
...
+    subject: Optional[Node] = None,
+    predicate: Optional[Node] = RDF.value,
+    _object: Optional[Node] = None,
+    default: Optional[Any] = None,
+    _any: Optional[bool] = True,
+) -> Optional[Union[URIRef, Literal, BNode]]:
+    value = graph.value(subject=subject, predicate=predicate, object=_object, default=default, any=_any)
+    if value and not isinstance(value, (URIRef, Literal, BNode)):
+        logger.append(
+            f"Warning: Object should be of type BNode, Literal or URIRef, but is {type(value).__name__}. "
```
armintaenzertng, line 142:
```suggestion
            f"Warning: Node should be of type BNode, Literal or URIRef, but is {type(value).__name__}. "
```

src/spdx/parser/rdf/graph_parsing_functions.py
```diff
@@ -102,3 +102,44 @@ def remove_prefix(string: str, prefix: str) -> str:
...
+    predicate: Optional[Node] = None,
+    _object: Optional[Node] = None,
+) -> Tuple[URIRef, Node, Union[BNode, Literal, URIRef]]:
+    # this is a helper method to enforce some rdf types for the triples to bo compatible with the
+    # code that follows
+    for s, p, o in graph.triples((subject, predicate, _object)):
+        if not isinstance(s, (BNode, URIRef)):
+            logger.append(
+                f"Warning: Subject should be of type BNode or URIRef, but is {type(s).__name__}. "
+                f"This might lead to a failure."
```
armintaenzertng, line 120: We might want to include the actual node, else it will be hard to discern what actually caused the warning.

armintaenzertng (commented):
Some comments on the new triples functions

src/spdx/parser/rdf/graph_parsing_functions.py
meretp: Good catch

src/spdx/parser/rdf/graph_parsing_functions.py
meretp: done

src/spdx/parser/rdf/graph_parsing_functions.py
meretp: fixed

src/spdx/parser/rdf/graph_parsing_functions.py
meretp: Done

armintaenzertng (approved):
Looks good now :)

armintaenzertng (approved):

PR #566, opened 2023-04-06
[issue-427] add bumping of ExternalPackageRef

armintaenzertng:
part of #427

### DIFF
--- a/src/spdx3/bump_from_spdx2/package.py
+++ b/src/spdx3/bump_from_spdx2/package.py
@@ -1,14 +1,19 @@
 # SPDX-FileCopyrightText: 2023 spdx contributors
 #
 # SPDX-License-Identifier: Apache-2.0
+from typing import Union
+
 from spdx3.bump_from_spdx2.actor import bump_actor
 from spdx3.bump_from_spdx2.bump_utils import handle_no_assertion_or_none
 from spdx3.bump_from_spdx2.checksum import bump_checksum
 from spdx3.bump_from_spdx2.message import print_missing_conversion
 from spdx3.model.creation_information import CreationInformation
+from spdx3.model.external_identifier import ExternalIdentifier, ExternalIdentifierType
+from spdx3.model.external_reference import ExternalReference, ExternalReferenceType
 from spdx3.model.software.package import Package
 from spdx3.model.software.software_purpose import SoftwarePurpose
 from spdx3.payload import Payload
+from spdx.model.package import ExternalPackageRef
 from spdx.model.package import Package as Spdx2_Package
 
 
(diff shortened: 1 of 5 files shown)


src/spdx3/bump_from_spdx2/package.py
```diff
@@ -57,10 +71,42 @@ def bump_package(
...
+    "gitoid": ExternalIdentifierType.GITOID,
+}
+
+
+def bump_external_package_ref(spdx2_external_ref: ExternalPackageRef) -> Union[ExternalReference, ExternalIdentifier]:
+    type = spdx2_external_ref.reference_type
+    locator = spdx2_external_ref.locator
+    comment = spdx2_external_ref.comment
+
+    id_or_ref = external_ref_map[type]
```
meretp, line 107: In SPDX 2.3 there's also `OTHER` as a category, where `reference_type` could be any string, this will lead to a `KeyError` here.

src/spdx3/bump_from_spdx2/package.py
meretp, line 1: From the migration guide:

> If there is a single ExternalRef of type purl, place that in the packageUrl property.

This should probably also be implemented.

tests/spdx3/bump/test_package_bump.py
meretp, line 1: I would add a test for `bump_external_package_ref` and add all possible `ExternalPackageRefs` from SPDX 2.3 for completeness.

meretp (changes requested):
Thanks for adding this. I think that there are two "bumps" missing, see my comments below.

tests/spdx3/bump/test_package_bump.py
armintaenzertng: After thinking about the implementation of such a test, I feel it would just be a duplication of the `bump_external_package_ref` method, which would not make much sense to me. Do you have a better idea?

src/spdx3/bump_from_spdx2/package.py
armintaenzertng: done

tests/spdx3/bump/test_package_bump.py
meretp: At the moment, yes it would be some kind of duplication but as soon as anyone makes any changes to this method the test would indicate if there is a problem. Especially with these different (non-trvial) mappings from `reference_type` to identifier/ locator and the newly raised `NotImplementedError`. But I don't think that this is merge-blocking, just a nice to have. I am also fine with postponing this to when all open question according to this bump are solved.

meretp (approved):
LGTM! And as I saw in the migration guide, you also added a comment about the probably missing comment when placing the ExternalRef in the purl property, very good catch!

tests/spdx3/bump/test_package_bump.py
armintaenzertng: Let's think about it later, then :)

PR #567, opened 2023-04-06
[issue-394] add tests for tag_value writer

### LINKED ISSUE #394: Add tests for tag-value writer

meretp:
fixes #394


armintaenzertng (commented):
Thanks for the effort, much appreciated! :)
However, I think it would be in our best interest to replace all of the value strings with references to the fixture values, at least where possible.
For example, `"FileName: ./fileName.py\n"` would become `f"FileName: {file_fixture().name}\n"`

meretp:
You are absolutely right, I replaced most strings. I didn't replace date strings and other strings that have a bit more logic than just a reference to the fixture value, please have another look.

tests/spdx/writer/tagvalue/test_file_writer.py
```diff
@@ -0,0 +1,38 @@
...
+        [
+            call("## File Information\n"),
+            call(f"FileName: {spdx_file.name}\n"),
+            call(f"SPDXID: {spdx_file.spdx_id}\n"),
+            call(f"FileType: {spdx_file.file_types[0].name}\n"),
+            call("FileChecksum: SHA1: 71c4025dd9897b364f3ebbb42c484ff43d00791c\n"),
+            call("LicenseConcluded: MIT AND GPL-2.0-only\n"),
+            call("LicenseInfoInFile: MIT\n"),
+            call("LicenseInfoInFile: GPL-2.0-only\n"),
+            call("LicenseInfoInFile: NOASSERTION\n"),
```
armintaenzertng, line 30: You missed a few here? (package and snippet have their license info referenced, too)

tests/spdx/writer/tagvalue/test_annotation_writer.py
```diff
@@ -0,0 +1,28 @@
...
+    annotation = annotation_fixture()
+
+    mock: MagicMock = mock_open()
+    with patch(f"{__name__}.open", mock, create=True):
+        with open("foo", "w") as file:
+            write_annotation(annotation, file)
+
+    mock.assert_called_once_with("foo", "w")
+    handle = mock()
+    handle.write.assert_has_calls(
```
armintaenzertng, line 20: This method allows other calls before or after the given list of calls. I haven't found a way to check that there are no such calls but maybe this isn't too important to check right now, anyway.

armintaenzertng (approved):
Thanks, no big issues from my side!

tests/spdx/writer/tagvalue/test_annotation_writer.py
meretp: Yes, another option would be `assert_has_calls` but this wouldn't check the order of calls, which is more important I think than the "risk" of calls before or after this statement.

PR #568, opened 2023-04-06
[issue-558] add test for optional feature to GitHub Action

### LINKED ISSUE #558: add graph visualization for SPDX documents
We want to add an optional feature to print the structure of a SPDX document as a graph. See the example below
image

meretp:
I decided to add the installation of the optional dependency and the test as additional steps to the GitHub Action so that we ensure on the one hand that the source installation is successful und all tests pass (resp. are skipped) and on the other hand also run the tests for the optional feature. 

addition to #562, fixes #558

Edit: I updated the test to only be skipped if `networkx` is not installed and install this package manually in the GitHub Action, as this is the only package needed for the tests. `pygraphviz` requires an installation of `graphviz` on the operating system. This seems to be more involving for Windows so I decided not to add this for the moment, as this is only an optional feature.

### DIFF
--- a/.github/workflows/install_and_test.yml
+++ b/.github/workflows/install_and_test.yml
@@ -33,3 +33,8 @@ jobs:
         run: pytest
       - name: Run CLI
         run: pyspdxtools -i ./tests/spdx/data/formats/SPDXJSONExample-v2.3.spdx.json
+
+      - name: Install optional dependencies
+        run: python -m pip install networkx
+      - name: Run tests for graph generation
+        run: pytest tests/spdx/test_graph_generation.py
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,8 +34,10 @@ Here's the process to make changes to the codebase:
    git checkout -b fix-or-improve-something
    python -m venv ./venv
    ./venv/bin/activate
-   pip install -e .
+   pip install -e ".[development]"
    ```
+   Note: By using the group `[development]` for the installation, all dependencies (including optional ones) will be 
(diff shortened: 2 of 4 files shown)


.github/workflows/install_and_test.yml
```diff
@@ -33,3 +33,8 @@ jobs:
         run: pytest
       - name: Run CLI
         run: pyspdxtools -i ./tests/spdx/data/formats/SPDXJSONExample-v2.3.spdx.json
+
+      - name: Install optional dependencies
+        run: python -m pip install networkx
+      - name: Run tests for graph generation
+        run: pytest tests/spdx/test_graph_generation.py
```
armintaenzertng, line 40: Can't we just install the optional dependencies together with the rest under "Installation"? Then we wouldn't need to call pytest again here.
By the way, is there a way to make sure that these tests actually are not skipped in the workflow?

armintaenzertng (commented):
Good catch!
I have a question, though.

.github/workflows/install_and_test.yml
meretp: One approach to enforce the execution of the tests would be to set an environment variable like `PYTEST_DISALLOW_SKIP` and set this in the GitHub Action. Or we could check the output of `pytest` if the tests were executed. As all of this seems rather complicated, I think we can postpone this issue and stick to the current GitHub Action.

meretp: https://github.com/spdx/tools-python/issues/572

armintaenzertng (approved):
LGTM

PR #569, opened 2023-04-11
[issue-427] add purl special case with comment distinction

armintaenzertng:
The package URL migration become even more complex, [see here](https://docs.google.com/document/d/1-olHRnX1CssUS67Psv_sAq9Vd-pc81HF8MM0hA7M0hg/edit#heading=h.r889hoh0obeu).

That is, a purl is converted to the new SPDX3 packageUrl property iff it is the only purl and has no comment.


meretp (approved):
LGTM

PR #571, opened 2023-04-11
[issue-570] change fixture dates

### LINKED ISSUE #570: Make fixture dates unique
The test data should not have the same datetime for different fields to ensure that they don't get mixed up unnoticed.

armintaenzertng:
fixes #570


meretp (approved):
lgtm

PR #574, opened 2023-04-12
[issue-573] unify `parse_from_file` tests for all formats

### LINKED ISSUE #573: Add yaml and xml parser tests
I noticed that there is no yaml and xml analogue for the [json parser test package](https://github.com/spdx/tools-python/tree/main/tests/spdx/parser/json). Let's add them.

armintaenzertng:
Fixes #573.
I went a little further and unified the general parsing tests for all formats into a single test class, also adding the v2.2 tag-value example.


tests/spdx/parser/all_formats/test_parse_from_file.py
```diff
@@ -0,0 +1,61 @@
...
+        (xml_parser, "XML", ".xml"),
+        (yaml_parser, "YAML", ".yaml"),
+        (rdf_parser, "Rdf", ".rdf.xml"),
+        (tagvalue_parser, "Tag", ""),
+    ],
+)
+class TestParseFromFile:
+    def test_parse_from_file_not_found(self, parser, format_name, extension):
+        with pytest.raises(FileNotFoundError) as err:
+            wrong_file_path = os.path.join(os.path.dirname(__file__), "hnjfkjsedhnflsiafg.json")
```
meretp, line 30: You could use `extension` here. Doesn't make a difference but feels more consistent regarding the different parsers

tests/spdx/data/formats/SPDXTagExample-v2.2.spdx
meretp, line 1: I am wondering if we really need the folder `formats`. Right now this gives an additional depth, that is not really  needed.

meretp (approved):
Thanks! Two comments from my side, that are rather optional.

tests/spdx/parser/all_formats/test_parse_from_file.py
armintaenzertng: This feels necessary and unnecessary at the same time :D
I'll add it

Issue #402, opened 2023-01-03
Make json and tag-value writer more consistent


armintaenzertng:
The two writers have been written by two different persons, thus sharing some common code which can be extracted and sporting slightly different API, which can be unified.

meretp:
Additionally the json/yaml/xml writers optionally validate the given file whereas the tag-value writer writes the document without validation.

armintaenzertng:
All writers are now consistent.

PR #576, opened 2023-04-12
[issue-402] change `write_document()` to `write_document_to_file()`

armintaenzertng:
part of #402 
This will make it consistent with the other functions of other formats.


meretp (approved):
LGTM

PR #577, opened 2023-04-12
[issue-540] change DESCRIBES validation

armintaenzertng:
The new logic is now: 
A `DESCRIBES` relationship is not mandatory if there is only a single SPDX element present which is a package.


meretp (approved):
Thanks! I stumbled a bit over the wording of the validation message since negations of cardinalities are a bit tricky but I don't have a better suggestion.

PR #582, opened 2023-04-13
[issue-413] add workflow for circle conversion integration test

### LINKED ISSUE #413: Add workflow for integration tests
Add a GitHub Actions workflow using the CLI tool to test end-to-end functionality of the project.
For example:
`origin.json` -> `first_conversion.json` -> `second_conversion.json`
-> check that `first_conversion.json` equals `second_conversion.json` (easy, because they should equal exactly)
`origin.json` -> `converted.xml` -> `converted.yaml` (-> `converted.tag` -> `converted.rdf`) -> `converted.json`
-> check that `origin.json` equals `converted.json`(harder, because they should only be equal up to ordering)

armintaenzertng:
fixes #413

### DIFF
--- a/.github/workflows/integration_test.yml
+++ b/.github/workflows/integration_test.yml
@@ -1,0 +1,43 @@
+name: Circle conversion test
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  install_and_test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.11
+      - name: Installation
+        run: |
+          python -m pip install --upgrade pip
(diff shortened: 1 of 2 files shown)


goneall:
I don't know if this helps - but here's a pointer to how the Java library handles this issue: 

https://github.com/spdx/Spdx-Java-Library/blob/da021dffa3c2376f2aa2654863fb000be2aa7d2a/src/main/java/org/spdx/library/model/license/DisjunctiveLicenseSet.java#L106

We basically override the equals and hashcode methods to "flatten" the set and do a compare.  Equals is overridden in most, if not all, of the license related classes.

maxhbr (approved):
LGTM

armintaenzertng:
After discussing this again, a quick fix for the problem is to simply exclude `AND` and `OR` license expressions from the test file. With this, the core of the test is still intact and we don't lose very much test coverage.

fholger (approved):

maxhbr (approved):

PR #584, opened 2023-04-14
[issue_583] fix type hint

### LINKED ISSUE #583: Incorrect type hint for field `copyright_text` in `Snippet`
The type hint for `copyright_text` in the constructor of a `Snippet` is only `Optional[str]` where it should be `Optional[Union[str, SpdxNoAssertion, SpdxNone]]`.

meretp:
fixes #583


armintaenzertng (approved):
Good catch! :)

Issue #426, opened 2023-01-17
Prototype SPDX3 model


meretp:
Core: 
- Classes
  - [x] Annotation
  - [ ] AnonymousPayload (insufficiently defined)
  - [x] Artifact
  - [x] Bom
  - [x] Bundle
  - [x] CreationInformation
  - [x] Element
  - [ ] ExtensionClass (insufficiently defined, see field extension in Element)
  - [x] ExternalIdentifier
  - [x] ExternalMap
  - [x] ExternalReference 
  - [x] Hash
  - [x] Identity (insufficiently defined, see also Entity)
  - [x] IntegrityMethod
  - [x] MediaType (no dedicated class as it only contains one string)
  - [x] NamespaceMap
  - [x] Organization (insufficiently defined, see also Entity)
  - [ ] Payload (insufficiently defined)
  - [x] Person (insufficiently defined, see also Entity)
  - [x] ProfileIdentifier
  - [x] Relationship
  - [x] SemVer (no dedicated class as we are using [semantic_version](https://pypi.org/project/semantic-version/)- a python package)
  - [x] SpdxCollection (ElementCollection in png)
  - [x] SpdxDocument
  - [x] Tool (insufficiently defined, see also Entity)

- Vocabularies
  - [x] ActorType (probably deprecated, replaced by Entity/Identity subclasses)
  - [x] AnnotationType
  - [x] ExternalIdentifierType
  - [x] ExternalReferenceType
  - [x] HashAlgorithm
  - [x] RelationshipCompleteness (contradicting definitions, known should be complete, implementation based on md-files)
  - [x] RelationshipType

Software:
- Classes
  - [x] File
  - [x] Package
  - [x] Snippet
  - [x] Sbom
  - [x] SoftwareDependencyRelationship (only in png)
 
- Vocabularies
  - [x] SoftwarePurpose

commit: a2e965924ba90c6e111e9631d836039819bea572

armintaenzertng:
The prototype is implemented, now the model has to be kept up-to-date, see #717

PR #585, opened 2023-04-17
[issue-426] add `source_info` to `Package`

armintaenzertng:
part of #426 and #427


meretp (changes requested):
Thanks! Could you please also update the `README.md` according to the commit these changes are based on.

meretp (approved):
Thanks!

PR #588, opened 2023-04-17
[issue-586] fix CLI

### LINKED ISSUE #586: CLI raises error if no arguments provided
If we only call `pyspdxtools` without any argument an `AttributeError` is raised. We should instead give a hint on how to correctly use the CLI and which arguments are possible.

meretp:
With this PR we make infile a required argument to display a help message when no arguments are given.
I also added some test to check that the cli fails for invalid arguments and passes for correct arguments.

fixes #586

### DIFF
--- a/tests/spdx/test_cli.py
+++ b/tests/spdx/test_cli.py
@@ -1,0 +1,45 @@
+import os
+
+import pytest
+from click.testing import CliRunner
+
+from spdx.clitools.pyspdxtools import main
+
+
+@pytest.mark.parametrize(
+    "options",
+    [
+        ("--infile", os.path.join(os.path.dirname(__file__), "data/SPDXJSONExample-v2.3.spdx.json")),
+        ("-i", os.path.join(os.path.dirname(__file__), "data/SPDXJSONExample-v2.3.spdx.json"), "--novalidation"),
+        (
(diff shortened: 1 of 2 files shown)


armintaenzertng (dismissed):
Thanks for the addition :)

tests/spdx/test_cli.py
```diff
@@ -0,0 +1,45 @@
...
+
+@pytest.mark.parametrize(
+    "options",
+    [
+        (),
+        ("-i", os.path.join(os.path.dirname(__file__), "data/SPDXJSONExample-v2.3.spdx.json"), "--version"),
+        ("-i", os.path.join(os.path.dirname(__file__), "data/SPDXJSONExample-v2.3.spdx.json"), "-o"),
+    ],
+)
+def test_cli_with_system_exit(options):
```
armintaenzertng, line 40: Sorry, one other thing; every CLI call has a system exit, the important part here is the exit code:
```suggestion
def test_cli_with_system_exit_code_2(options):
```

armintaenzertng: maybe also adapt the other test name to be consistent

armintaenzertng (approved):

PR #590, opened 2023-04-17
[issue_589] fix json parser: process fields that can be "NOASSERTION"…

### LINKED ISSUE #589: JSON parser: homepage of package not parsed correctly
The `homepage` field of a package could be `"NOASSERTION"` or `"NONE"`.
When parsing a file like the one below
```
{
    "spdxVersion": "SPDX-2.3",
    "documentNamespace": "https://true",
    "creationInfo": {
        "creators": [
            "Person: Mee (mee)"
        ],
        "created": "2023-04-17T12:34:39Z",
        "licenseListVersion": "3.20"
    },
    "dataLicense": "CC0-1.0",
    "SPDXID": "SPDXRef-DOCUMENT",
    "name": "Test 2",
    "packages": [
        {
            "SPDXID": "SPDXRef-package",
            "name": "Test",
            "downloadLocation": "NOASSERTION",
            "homepage": "NOASSERTION"
        }]

}
``` 
a `ValidationError` is raised. This originates from the JSON parser which doesn't process the field correctly.

meretp:
… or "NONE" correctly
The issue is only related to the homepage field, while working on the fix, I noticed that the issue also applies to other fields.
 
fixes #589

### DIFF
--- a/tests/spdx/parser/jsonlikedict/test_package_parser.py
+++ b/tests/spdx/parser/jsonlikedict/test_package_parser.py
@@ -11,12 +11,66 @@ from spdx.model.actor import Actor, ActorType
 from spdx.model.checksum import Checksum, ChecksumAlgorithm
 from spdx.model.package import ExternalPackageRef, ExternalPackageRefCategory, PackagePurpose, PackageVerificationCode
 from spdx.model.spdx_no_assertion import SpdxNoAssertion
+from spdx.model.spdx_none import SpdxNone
 from spdx.parser.error import SPDXParsingError
 from spdx.parser.jsonlikedict.dict_parsing_functions import parse_list_of_elements
 from spdx.parser.jsonlikedict.package_parser import PackageParser
 
 
-def test_parse_package():
+@pytest.mark.parametrize(
+    "homepage, expected_homepage, download_location, expected_download_location, "
+    "copyright_text, expected_copyright_text, originator, expected_originator, supplier, expected_supplier",
+    [
+        (
+            "http://ftp.gnu.org/gnu/glibc",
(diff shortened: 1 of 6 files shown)


tests/spdx/parser/jsonlikedict/test_package_parser.py
```diff
@@ -11,12 +11,66 @@
...
+            "NONE",
+            SpdxNone(),
+            "http://ftp.gnu.org/gnu/glibc/glibc-ports-2.15.tar.gz",
+            "http://ftp.gnu.org/gnu/glibc/glibc-ports-2.15.tar.gz",
+            "NOASSERTION",
+            SpdxNoAssertion(),
+            "NOASSERTION",
+            SpdxNoAssertion(),
+            "Person: Jane Doe (jane.doe@example.com)",
+            Actor(ActorType.PERSON, "Jane Doe", "jane.doe@example.com"),
```
armintaenzertng, line 58: This is kind of hard to comprehend due to the scrambling of normal, none and no assertion cases. I get that this might make the test cases more diverse, but I don't think it is really worth it compared to just having test cases where all values are normal values or None/NONE etc.

armintaenzertng (approved):
Good catch! :)
I have a small remark that I don't feel very strongly about.

tests/spdx/parser/jsonlikedict/test_package_parser.py
meretp: Yes, my intention was to make the test cases more diverse. I think both options doesn't really make a difference and as they are also just test cases  I would stick to the current implementation and merge it like that.

PR #595, opened 2023-04-19
[issue_592] add top-level package

### LINKED ISSUE #592: `src/common` should be installed inside package
When running `pip install .` from the latest `main` branch, I end up with `common`, `spdx` and the `spdx_tools-0.7.2.dev397+gd9a2e01.dist-info` directories inside my `venv/lib/python3.10/site-packages` directory.

This does not seem right, as `common` should rather be part of the `spdx` distribution itself to avoid confusion.

meretp:
This PR adds the packagee `spdx_tools` at top-level. With this the subpackages `spdx` and `common` are bundled under the package `spdx_tools` when installing the tools. We descided to use this structure to be able to include the `spdx3` package next to the `spdx` package which is currently [prototyped](https://github.com/spdx/tools-python/tree/prototype-spdx-3.0) and will be added once SPDX 3.0 is published. 

I will also update the migration guide according to the changed imports. 

As this is a rather big refactoring please look at it carefully and tell me if I missed to adapt anything. 

fixes #592


armintaenzertng (approved):
Thanks, lgtm

Issue #593, opened 2023-04-18
Example missing imports


stefan6419846:
I just started evaluating this project for generating SPDX files manually, but had some trouble to finally generate a custom document:

* The example at https://github.com/spdx/tools-python#example omits all imports.
* The example starts with parsing an existing file. As documentation is sparse anyway, this requires me to do quite some extra research on how to create a document itself manually.
* Instead of exposing the corresponding classes from the submodules of `spdx.models` inside `spdx/models/__init__.py`, I am required to import them from the corresponding submodules, which seems like some unnecessary overhead.

armintaenzertng:
Thanks for the detailed feedback, Stefan, much appreciated! :)
As a first quick fix I opened a PR to include the imports in the readme example. I'll work on another example that demonstrates creating documents from scratch.

armintaenzertng:
@stefan6419846, I have [created a new example file](https://github.com/spdx/tools-python/pull/598) that demonstrates how to create an SPDX document from scratch with this library. Please have a look and tell me if you find that helpful or where you wish for more information. :)

stefan6419846:
Thanks, this is much appreciated to lower the burden of getting started with plain library usage.

Doing some further research in the package itself, I managed to write a similar example in the meantime (with the help of the actual specification), but good documentation does not hurt.

armintaenzertng:
That's great to here! Please reach out if there are any more road blocks you encounter!

PR #597, opened 2023-04-19
[issue-593] expose model classes in model package init

armintaenzertng:
I had to exclude the `__init__.py` from isort and flake8 because the order of the imports matters in that case.

part of #593

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,4 @@
 [flake8]
 max-line-length = 119
-exclude = src/spdx_tools/spdx/parser/tagvalue/parsetab.py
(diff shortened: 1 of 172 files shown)


meretp (approved):
Thanks! I think this is real improvement concerning readability. I was wondering if we might want to use this for other classes outside of `spdx_tools.spdx.model`, for example the parser class in `spdx_tools.spdx.parser.jsonlikedict`, what do you think?

armintaenzertng:
I also think that this approach can be applied to other packages as well, but it probably has the most impact for the model package, as that is the one that will be most widely used by users, I imagine.

PR #600, opened 2023-04-20
[issue-593] expose model classes at model package level

armintaenzertng:
analogous to #597, this time for the SPDX3 model packages.

### DIFF
--- a/tests/spdx3/model/test_relationship.py
+++ b/tests/spdx3/model/test_relationship.py
@@ -5,10 +5,10 @@ from unittest import mock
 
 import pytest
 
-from spdx_tools.spdx3.model.relationship import Relationship, RelationshipCompleteness, RelationshipType
+from spdx_tools.spdx3.model import Relationship, RelationshipCompleteness, RelationshipType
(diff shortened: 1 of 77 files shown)


tests/spdx3/model/test_relationship.py
```diff
@@ -5,7 +5,7 @@
 
 import pytest
 
-from spdx_tools.spdx3.model.relationship import Relationship, RelationshipCompleteness, RelationshipType
+from spdx_tools.spdx3.model import Relationship, RelationshipCompleteness, RelationshipType
 
 
 @mock.patch("spdx_tools.spdx3.model.creation_information.CreationInformation", autospec=True)
```
meretp, line 11: Nit: You could also adapt the paths in the mocks.

meretp (approved):
LGTM, just one small remark from my side.

tests/spdx3/model/test_relationship.py
armintaenzertng: Good point, added

PR #603, opened 2023-04-20
refactor bumping modules

armintaenzertng:
While going through the bumping modules, I had the urge to utilize the new black formatting a bit more and make the assignment of SPDX 2 to 3 properties a bit more comprehensible (hopefully). Now, for properties that just carry over 1-to-1, you don't have to look around in the module anymore to find the initialization of these variables.


meretp (approved):
LGTM, nice improvement concering readability and simplicity :+1:

PR #604, opened 2023-04-20
[issue-426] add Dataset class

armintaenzertng:
part of #426

### DIFF
--- a/src/spdx_tools/spdx3/model/dataset/dataset.py
+++ b/src/spdx_tools/spdx3/model/dataset/dataset.py
@@ -1,0 +1,59 @@
+#  SPDX-FileCopyrightText: 2023 spdx contributors
+#
+#  SPDX-License-Identifier: Apache-2.0
+from dataclasses import field
+from enum import Enum, auto
+from typing import Dict, List, Optional
+
+from spdx_tools.common.typing.dataclass_with_properties import dataclass_with_properties
+from spdx_tools.common.typing.type_checks import check_types_and_set_values
+
+
+class ConfidentialityLevelType(Enum):
+    RED = auto()
+    AMBER = auto()
+    GREEN = auto()
+    CLEAR = auto()
+
+
+class DatasetAvailabilityType(Enum):
+    DIRECT_DOWNLOAD = auto()
+    SCRAPING_SCRIPT = auto()
+    QUERY = auto()
+    CLICKTHROUGH = auto()
+    REGISTRATION = auto()
+
+
+@dataclass_with_properties
+class Dataset:
+    data_collection_process: Optional[str] = None
+    intended_use: Optional[str] = None
+    dataset_size: Optional[int] = None
+    dataset_noise: Optional[str] = None
+    data_preprocessing_steps: Optional[str] = None
+    sensors: Dict[str, Optional[str]] = field(default_factory=dict)
+    known_biases: Optional[str] = None
+    sensitive_personal_information: Optional[bool] = None
+    anonymization_method_used: List[str] = field(default_factory=list)
+    confidentiality_level: Optional[ConfidentialityLevelType] = None
+    dataset_update_mechanism: Optional[str] = None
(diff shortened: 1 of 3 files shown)


src/spdx_tools/spdx3/model/dataset/dataset.py
```diff
@@ -0,0 +1,56 @@
...
+    data_collection_process: Optional[str] = None
+    intended_use: Optional[str] = None
+    dataset_size: Optional[int] = None
+    dataset_noise: Optional[str] = None
+    data_preprocessing_steps: Optional[str] = None
+    sensors: Optional[Dict[str, str]] = None
+    known_biases: Optional[str] = None
+    sensitive_personal_information: Optional[bool] = None
+    anonymization_method_used: Optional[str] = None
+    confidentiality_level: Optional[ConfidentialityLevelType] = None
```
meretp, line 37: These 3 properties don't have a maxCount, so I would expect lists here.

src/spdx_tools/spdx3/model/dataset/dataset.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+@dataclass_with_properties
+class Dataset:
+    data_collection_process: Optional[str] = None
+    intended_use: Optional[str] = None
+    dataset_size: Optional[int] = None
+    dataset_noise: Optional[str] = None
+    data_preprocessing_steps: Optional[str] = None
+    sensors: Optional[Dict[str, str]] = None
```
meretp, line 33: This property doesn't have a maxCount neither, so I would expect a list here.

src/spdx_tools/spdx3/model/dataset/dataset.py
```diff
@@ -0,0 +1,56 @@
...
+
+from spdx_tools.common.typing.dataclass_with_properties import dataclass_with_properties
+from spdx_tools.common.typing.type_checks import check_types_and_set_values
+
+
+class ConfidentialityLevelType(Enum):
+    RED = auto()
+    AMBER = auto()
+    GREEN = auto()
+    CLEAR = auto()
```
meretp, line 15: This one is not listed in the markdown file, but probably an issue in the spdx-3-model?

src/spdx_tools/spdx3/model/dataset/dataset.py
meretp: I have one question concerning this: the markdown file for `DictionaryEntry` states that the property `value` is optional, how would we handle this? I would think `List[Dict[str, Optional[str]]]` would be the correct type hint.

meretp (changes requested):
Thanks! I have some comments/ remarks concerning the type hints, not sure if these are issues in the spec or in this PR.

src/spdx_tools/spdx3/model/dataset/dataset.py
armintaenzertng: Good point for `anonymization_method_used`; for the other two I [opened a PR](https://github.com/spdx/spdx-3-model/pull/219) as I think they actually should have a `maxCount`.

src/spdx_tools/spdx3/model/dataset/dataset.py
armintaenzertng: Yes, I opened an issue for this.

src/spdx_tools/spdx3/model/dataset/dataset.py
meretp: The only important difference would be the need for unique keys. So I don't know if this makes any sense, but from the definition in the spdx-3-model we could have two sensors with the same key, don't we?

armintaenzertng: I believe this is still debated whether this is actually wanted in the spdx-3-model... If yes, then you are right, we have a problem

armintaenzertng: Removed the `Optional`

meretp (approved):
lgtm

Issue #602, opened 2023-04-20
Problems when converting JSON 2.3 to 3.0


meretp:
I tried to use the cli to convert the test file SPDXJSONExample-v2.3.spdx.json to a SPDX 3.0 file but got some uncatched errors.
The errors were raised due to the following fields: 

- [x] If no `originator`  is  set, the bumping of a package fails
- [x] the relationshiptype `SPECIFICATION_FOR` 
- [x] the relationship to `NOASSERTION`

meretp:
gixed with #605 and #620

PR #605, opened 2023-04-20
[issue_602] fix package bump 

meretp:
If originator is NOASSERTION or not provided we don't need to call `bump_actor`.

subtask of #602

### DIFF
--- a/tests/spdx3/bump/test_package_bump.py
+++ b/tests/spdx3/bump/test_package_bump.py
@@ -3,26 +3,38 @@
 # SPDX-License-Identifier: Apache-2.0
 from unittest import TestCase, mock
 
+import pytest
+
 from spdx_tools.spdx3.bump_from_spdx2.package import bump_package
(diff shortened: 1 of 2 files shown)


tests/spdx3/bump/test_package_bump.py
```diff
@@ -3,26 +3,36 @@
...
 from spdx_tools.spdx.model.package import Package as Spdx2_Package
-from tests.spdx.fixtures import package_fixture
+from tests.spdx.fixtures import actor_fixture, package_fixture
 
 
+@pytest.mark.parametrize(
+    "originator, expected_originator",
+    [
+        (actor_fixture(name="originatorName"), "https://doc.namespace#SPDXRef-Actor-originatorName-some@mail.com"),
+        (None, None),
```
armintaenzertng, line 21: Let's also insert the NOASSERTION case while we are it.

meretp: Done

PR #606, opened 2023-04-20
make lists non-optional in model classes

armintaenzertng:
I noticed that -- contrary to the SPDX2 model -- lists were optional in the new SPDX3 model.
We should keep this consistent, so I made them non-optional again. Now, an empty list is the analogue of no list at all.


meretp (approved):
lgtm

PR #607, opened 2023-04-24
make example files valid

meretp:
I noticed that these two example files are marked as invalid. The changed files align with the examples given in the spdx-spec repo. As far as I understand the spec we only add the documents namespace to `referenceType` in `RDF` format, not in any other. So I think the examples were simply incorrect and there is no need to adapt the code to accept these kind of `referenceType`.


armintaenzertng (approved):
It certainly does not make sense to define "purl" in the local namespace.

We should also open a PR against the spec.

PR #610, opened 2023-04-24
[issue-426] make Dataset a subclass of Package

armintaenzertng:
Oversight during #604: Dataset is a subclass of package.


meretp (approved):
lgtm

PR #616, opened 2023-04-25
Issue 615 fix rdf parser

### LINKED ISSUE #615: RDF Parser fails with AttributeError if CreationInfo has bugs
When trying to parse a rdf file that has some bugs in the CreationInfo, e.g. a missing `created` tag the application fails with an uncatched`AttributeError`. This should raise an `SPDXParsingError`.

meretp:
This PR fixes two things:
- If there is any error while parsing the creation info  the parsing process will be canceled as the following code tries to access `creation_info.doc_namespace`. This lead to an uncaught `AttributeError`. 
- If there are no creators in the given input file an `SPDXParsingError` will be raised as this is not valid.

I added two negative tests for these cases. They seem to be pretty similar but they test different things. I added the `invalid_creation_info_with_snippet` to verify that there is no uncaught error whereas the other test simply tests that a file without creators will raise an `SPDXParsingError`. 

fixes #615


armintaenzertng (approved):
LGTM

PR #617, opened 2023-04-25
Make pytest not skip 'test_build' folders

armintaenzertng:
Per default, pytest skips over folder containing "build". This overwrites that default.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,3 +62,6 @@ include = "(^/src/.*.py|^/tests/.*.py)"
 profile = "black"
 line_length = 119
 skip = ["__init__.py"]
+
+[tool.pytest.ini_options]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -62,3 +62,6 @@ include = "(^/src/.*.py|^/tests/.*.py)"
 profile = "black"
 line_length = 119
 skip = ["__init__.py"]
+
+[tool.pytest.ini_options]
+norecursedirs = []  # overwrite the default because per default the test_build folder is skipped
```
meretp, line 67:
```suggestion
norecursedirs = []  # overwrite the default to not skip tests/build folder which is needed in spdx3 prototype
```

armintaenzertng: changed. I omitted the "prototype" as we will probably forget to erase that later when it is not a mere prototype anymore

meretp (approved):
lgtm

Issue #434, opened 2023-01-19
implement a debug writer for the protoype model 3.0


meretp:
To be able to track the current state of the protoype we want to have a human-readable output of the implemented model.

meretp:
started with #436 
additions to the model should also implement a debug writer, I would leave this issue open to not forget to implement the debug writer.

PR #619, opened 2023-04-26
[issue-434] refactor debug writers

armintaenzertng:
Tested with `AIPackage`

part of #434

### DIFF
--- a/src/spdx_tools/spdx3/writer/console/console.py
+++ b/src/spdx_tools/spdx3/writer/console/console.py
@@ -1,21 +1,26 @@
 # SPDX-FileCopyrightText: 2023 spdx contributors
 #
 # SPDX-License-Identifier: Apache-2.0
-
+from enum import Enum
 from typing import Optional, TextIO, Union
 
 
-def write_value(tag: str, value: Optional[Union[bool, str]], out: TextIO, indent: bool = False):
+def write_value(tag: str, value: Optional[Union[bool, str, dict, list, Enum]], out: TextIO, indent: bool = False):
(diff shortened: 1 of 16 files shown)


src/spdx_tools/spdx3/writer/console/console.py
```diff
@@ -1,21 +1,29 @@
...
     and slightly adapted to make indentation of output possible."""
     if not value:
         return
-    if indent:
-        out.write(f"\t{tag}: {value}\n")
-    else:
-        out.write(f"{tag}: {value}\n")
+    if isinstance(value, dict):
+        for key, val in value.items():
+            write_and_possibly_indent(f'{tag}: "{key}": "{val}"', indent, out)
```
meretp, line 15: You put `"` here to separate the dict output from the other tag-value combination, right? Probably more a matter of taste but I would rather print key-value pairs, so `({key}, {value})` here to avoid these `"`, but this is completely up to you.

armintaenzertng:
I felt like there was too much noise going on in the other writers to warrant the effort, but I had another look and will adapt some more.

src/spdx_tools/spdx3/writer/console/console.py
armintaenzertng: Sure, I don't have strong feelings on this one.

armintaenzertng: Do you think we should have it as a list of tuples then instead of every entry in a new line?

meretp: I don't think this makes any difference. Having a list of tuples would be consistent with the handling of lists, so yes, might be good.

meretp (approved):
LGTM

PR #620, opened 2023-04-26
[issue_602] update relationship bump

### LINKED ISSUE #602: Problems when converting JSON 2.3 to 3.0

meretp:
This PR is based on #609.  

I assumed that relationships with the same "from spdx_id"  and relationship_type should be merged. Sone conversions are not yet defined or the RelatoinshipType doesn't exist yet in the spdx-3-model and the prototype, I decided to print an info about missing conversions here and added a comment at the corresponding relationshiptypes. 

fixes #602

### DIFF
--- a/src/spdx_tools/spdx3/bump_from_spdx2/relationship.py
+++ b/src/spdx_tools/spdx3/bump_from_spdx2/relationship.py
@@ -187,9 +187,10 @@ def bump_relationships(
                 (relationship.from_element, relationship.relationship_type.name), []
             ).append(relationship)
 
-    for key, relationships in generated_relationships.items():
+    for relationships in generated_relationships.values():
         if len(relationships) > 1:
-            merge_relationships_and_add_to_payload(payload, relationships)
+            merged_relationship = merge_relationships_and_add_to_payload(relationships)
+            payload.add_element(merged_relationship)
         else:
             payload.add_element(relationships[0])
 
@@ -201,7 +202,7 @@ def bump_relationship(
     counter: int,
 ) -> Optional[Union[Relationship, SoftwareDependencyRelationship]]:
     swap_direction = False
-    completeness, to = determine_completeness(spdx2_relationship.related_spdx_element_id)
+    completeness, to = determine_completeness_and_to(spdx2_relationship.related_spdx_element_id)
     spdx_id = "#".join([document_namespace, f"SPDXRef-Relationship-{counter}"])
     parameters_for_bump = relationship_mapping[spdx2_relationship.relationship_type]
     if parameters_for_bump is None:
@@ -210,9 +211,6 @@ def bump_relationship(
     base_relationship = parameters_for_bump[0]
     relationship_type = parameters_for_bump[1]
     if not base_relationship:
-        scope = parameters_for_bump[2]
-        software_linkage = parameters_for_bump[3]
-        conditionality = parameters_for_bump[4]
         from_element = spdx2_relationship.spdx_element_id
 
         return SoftwareDependencyRelationship(
@@ -223,12 +221,12 @@ def bump_relationship(
             relationship_type,
             comment=spdx2_relationship.comment,
             completeness=completeness,
-            scope=scope,
-            software_linkage=software_linkage,
-            conditionality=conditionality,
+            scope=(parameters_for_bump[2]),
+            software_linkage=(parameters_for_bump[3]),
+            conditionality=(parameters_for_bump[4]),
         )
 
-    if base_relationship and (len(parameters_for_bump) == 3):
+    if len(parameters_for_bump) == 3:
         swap_direction = parameters_for_bump[2]
     if swap_direction:
         if not to:
@@ -250,7 +248,7 @@ def bump_relationship(
     )
 
 
-def determine_completeness(
+def determine_completeness_and_to(
     related_spdx_element_id: Union[str, SpdxNone, SpdxNoAssertion]
 ) -> Tuple[Optional[RelationshipCompleteness], List[str]]:
     if isinstance(related_spdx_element_id, SpdxNoAssertion):
@@ -265,7 +263,7 @@ def determine_completeness(
     return completeness, to
 
 
-def merge_relationships_and_add_to_payload(payload: Payload, relationships: List[Relationship]):
+def merge_relationships_and_add_to_payload(relationships: List[Relationship]) -> Relationship:
     merged_relationship = relationships[0]
     for relationship in relationships[1:]:
         merged_relationship.to += relationship.to
@@ -274,4 +272,4 @@ def merge_relationships_and_add_to_payload(payload: Payload, relationships: List
         if not merged_relationship.comment and relationship.comment:
             merged_relationship.comment = relationship.comment
 
-    payload.add_element(merged_relationship)
+    return merged_relationship
--- a/src/spdx_tools/spdx3/bump_from_spdx2/relationship.py
+++ b/src/spdx_tools/spdx3/bump_from_spdx2/relationship.py
@@ -1,6 +1,8 @@
 # SPDX-FileCopyrightText: 2023 spdx contributors
 #
 # SPDX-License-Identifier: Apache-2.0
+import logging
+import sys
 from typing import Dict, List, Optional, Tuple, Union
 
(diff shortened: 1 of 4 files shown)


src/spdx_tools/spdx3/bump_from_spdx2/relationship.py
```diff
@@ -1,74 +1,277 @@
...
+        relationship = bump_relationship(spdx2_relationship, creation_information, document_namespace, counter)
+        if relationship:
+            generated_relationships.setdefault(
+                (relationship.from_element, relationship.relationship_type.name), []
+            ).append(relationship)
 
-    spdx_id = "#".join([document_namespace, f"SPDXRef-Relationship-{counter}"])
+    for key, relationships in generated_relationships.items():
+        if len(relationships) > 1:
+            merge_relationships_and_add_to_payload(payload, relationships)
```
armintaenzertng, line 192: In the name of a single function not doing multiple things I suggest using only `merge_relationships(relationships)` and returning the new, merged relationship which can then be added here to the payload.

src/spdx_tools/spdx3/bump_from_spdx2/relationship.py
```diff
@@ -1,74 +1,277 @@
...
+
+
+def merge_relationships_and_add_to_payload(payload: Payload, relationships: List[Relationship]):
+    merged_relationship = relationships[0]
+    for relationship in relationships[1:]:
+        merged_relationship.to += relationship.to
+        if merged_relationship.comment and relationship.comment:
+            merged_relationship.comment += ", " + relationship.comment
+        if not merged_relationship.comment and relationship.comment:
+            merged_relationship.comment = relationship.comment
```
armintaenzertng, line 275: This overwrites the completeness of all relationship with that of the first one. I believe that if one of them has NOASSERTION, the resulting relationship should have, too.
There might be cases where the completeness statements contradict each other, we have to think how to deal with that.

src/spdx_tools/spdx3/bump_from_spdx2/relationship.py
```diff
@@ -1,74 +1,277 @@
...
             to,
             relationship_type,
             comment=spdx2_relationship.comment,
             completeness=completeness,
+            scope=scope,
+            software_linkage=software_linkage,
+            conditionality=conditionality,
         )
+
+    if base_relationship and (len(parameters_for_bump) == 3):
```
armintaenzertng, line 231: base_relationship is True here for sure, it is already tested in line 212.

src/spdx_tools/spdx3/bump_from_spdx2/relationship.py
```diff
@@ -1,74 +1,277 @@
...
+        swap_direction = parameters_for_bump[2]
```
armintaenzertng, line 232: I'm confused by the parameters; in line 213 you have `scope = parameters_for_bump[2]` ...?

armintaenzertng (commented):
Thanks for this!
I haven't looked at the mapping and the tests yet, but here are some comments already :)

src/spdx_tools/spdx3/bump_from_spdx2/relationship.py
```diff
@@ -1,74 +1,277 @@
...
     payload: Payload,
     creation_information: CreationInformation,
     document_namespace: str,
-    counter: int,
 ):
-    relationship_type, swap_direction = bump_relationship_type(spdx2_relationship.relationship_type)
+    generated_relationships: Dict[Tuple[str, str], List[Relationship]] = {}
+    for counter, spdx2_relationship in enumerate(spdx2_relationships):
+        relationship = bump_relationship(spdx2_relationship, creation_information, document_namespace, counter)
+        if relationship:
```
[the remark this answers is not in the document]
meretp: Yes this is due to the incomplete conversion table. Although, it seems that there are two relationships that won't result in a SPDX3 relationship if I understood the table correctly. But for now this just "handles" the open questions.

src/spdx_tools/spdx3/bump_from_spdx2/relationship.py
meretp: You are right, I also thought about contradicting relationships and I am still not sure if this merigng approach is a good idea. Maybe we should only merge relationships if they contain no additional information at all, so only merge "simple" from-relationship-type-to-spdx-element relationships. And as soon as there is a comment or `SpdxNoAssertion` or `SpdxNone` involved, don't merge anything...

src/spdx_tools/spdx3/bump_from_spdx2/relationship.py
meretp: This is some kind of trade-off between keeping the `relationship_mapping` as simple as possible and delaing with the different cases. So, all spdx2_relationships that can be swaooed will result in a "classic" relationship and all spdx2_relationships that result in a SoftwareDependencyRelationship need additional parameter for that. So the mapping is either spdx2_relationship -> (base_relationsihip, relationship_type, swap_direction) or spdx2_relationship -> (no_base_relationship, relationship_type, scope, software_linkage, conditionality). But we can also map each spdx2_relationship to a quintuple..

src/spdx_tools/spdx3/bump_from_spdx2/relationship.py
```diff
@@ -1,74 +1,277 @@
...
+    parameters_for_bump = relationship_mapping[spdx2_relationship.relationship_type]
+    if parameters_for_bump is None:
+        print_missing_conversion(spdx2_relationship.relationship_type.name, 0)
+        return
+    base_relationship = parameters_for_bump[0]
+    relationship_type = parameters_for_bump[1]
+    if not base_relationship:
+        scope = parameters_for_bump[2]
+        software_linkage = parameters_for_bump[3]
+        conditionality = parameters_for_bump[4]
```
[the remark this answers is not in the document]
meretp: agree

src/spdx_tools/spdx3/bump_from_spdx2/relationship.py
```diff
@@ -1,74 +1,277 @@
...
-    if spdx2_relationship_type.name.endswith("_FOR"):
-        relationship_type = spdx2_relationship_type.name.replace("_FOR", "")
-        return RelationshipType[relationship_type], False
-    return RelationshipType[spdx2_relationship_type.name], False
-    # if spdx2_relationship_type == Spdx2_RelationshipType.PATCH_APPLIED:
-    #     print_missing_conversion("RelationshipType.PATCH_APPLIED", 0)
-    #     return None
+def determine_completeness(
+    related_spdx_element_id: Union[str, SpdxNone, SpdxNoAssertion]
+) -> Tuple[Optional[RelationshipCompleteness], List[str]]:
```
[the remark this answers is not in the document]
meretp: changed

src/spdx_tools/spdx3/bump_from_spdx2/relationship.py
meretp: Agree, changed

src/spdx_tools/spdx3/bump_from_spdx2/relationship.py
```diff
@@ -1,74 +1,277 @@
...
+    generated_relationships: Dict[Tuple[str, str], List[Relationship]] = {}
+    for counter, spdx2_relationship in enumerate(spdx2_relationships):
+        relationship = bump_relationship(spdx2_relationship, creation_information, document_namespace, counter)
+        if relationship:
+            generated_relationships.setdefault(
+                (relationship.from_element, relationship.relationship_type.name), []
+            ).append(relationship)
 
-    spdx_id = "#".join([document_namespace, f"SPDXRef-Relationship-{counter}"])
+    for key, relationships in generated_relationships.items():
```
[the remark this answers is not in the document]
meretp: done

armintaenzertng (approved):

PR #621, opened 2023-04-26
Generate output instead of raising an error 

meretp:
To get the most information possible I replaced most places where an error is raised with an output to the console.


armintaenzertng (approved):
LGTM

PR #622, opened 2023-04-26
[issue_426] make name mandatory for package and file

meretp:
according to [#170](https://github.com/spdx/spdx-3-model/pull/170)


armintaenzertng (approved):
I had this on my todo list for a while now, thanks for taking care of it! :)

PR #627, opened 2023-05-02
Implemented fixtures for tests on the prototype spdx3

HarshvMahawar:
Analogous to the fixtures in spdx 2.x we also needed to use fixtures for the tests of the prototyped model.


meretp:
Thanks for your work on this @HarshvMahawar!

PR #628, opened 2023-05-02
Update model according to recent changes in spdx-3-model

meretp:
- adapt some field names to use singular
- add identifier_locator and issuing_authority to ExternalIdentifier

I went through all changes since the last tagged commit and updated the implementation.

### DIFF
--- a/src/spdx_tools/spdx3/model/ai/ai_package.py
+++ b/src/spdx_tools/spdx3/model/ai/ai_package.py
@@ -22,17 +22,17 @@ class SafetyRiskAssessmentType(Enum):
 @dataclass_with_properties
 class AIPackage(Package):
     energy_consumption: Optional[str] = None
-    standards_compliance: List[str] = field(default_factory=list)
-    limitations: Optional[str] = None
+    standard_compliance: List[str] = field(default_factory=list)
+    limitation: Optional[str] = None
     type_of_model: List[str] = field(default_factory=list)
     information_about_training: Optional[str] = None
     information_about_application: Optional[str] = None
-    hyperparameters: Dict[str, Optional[str]] = field(default_factory=dict)
-    data_preprocessing_steps: Optional[str] = None
-    model_explainability_mechanisms: Optional[str] = None
+    hyperparameter: Dict[str, Optional[str]] = field(default_factory=dict)
+    model_data_preprocessing: Optional[str] = None
+    model_explainability: Optional[str] = None
     sensitive_personal_information: Optional[bool] = None
-    metrics_decision_thresholds: Dict[str, Optional[str]] = field(default_factory=dict)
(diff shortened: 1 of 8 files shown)


src/spdx_tools/spdx3/model/ai/ai_package.py
```diff
@@ -22,17 +22,17 @@ class SafetyRiskAssessmentType(Enum):
...
-    hyperparameters: Dict[str, Optional[str]] = field(default_factory=dict)
-    data_preprocessing_steps: Optional[str] = None
-    model_explainability_mechanisms: Optional[str] = None
+    hyperparameter: Dict[str, Optional[str]] = field(default_factory=dict)
+    model_data_preprocessing: Optional[str] = None
+    model_explainability: Optional[str] = None
     sensitive_personal_information: Optional[bool] = None
-    metrics_decision_thresholds: Dict[str, Optional[str]] = field(default_factory=dict)
-    metrics: Dict[str, Optional[str]] = field(default_factory=dict)
+    metrics_decision_threshold: Dict[str, Optional[str]] = field(default_factory=dict)
```
armintaenzertng, line 34:
```suggestion
    metric_decision_threshold: Dict[str, Optional[str]] = field(default_factory=dict)
```
Also for other occurrences.

armintaenzertng (commented):
Thanks!
I found one inconsistency, probably not the last we'll encounter :O

src/spdx_tools/spdx3/model/ai/ai_package.py
meretp: Good catch, thanks! I updated this, rebased and as I needed to force-push anyway, I also squashed the commits, hope this is fine.

tests/spdx3/model/test_external_identifier.py
```diff
@@ -1,23 +1,31 @@
...
+        ExternalIdentifierType.CPE22,
+        "cpe:/o:canonical:ubuntu_linux:10.04:-:lts",
+        "This is a comment",
+        ["first locator", "second locator"],
+        "authority",
     )
     assert external_identifier.external_identifier_type == ExternalIdentifierType.CPE22
     assert external_identifier.identifier == "cpe:/o:canonical:ubuntu_linux:10.04:-:lts"
     assert external_identifier.comment == "This is a comment"
+    TestCase().assertCountEqual(external_identifier.identifier_locator, ["first locator", "second locator"])
```
armintaenzertng, line 22: I don't think Python messes up the order of lists between setting and retrieving it.
```suggestion
    assert external_identifier.identifier_locator == ["first locator", "second locator"]
```

armintaenzertng (approved):
Just a minor left-over comment, not merge-blocking from my side.

PR #633, opened 2023-05-09
Issue #426: update Relationship and RelationshipType

armintaenzertng:
adds `start_time` and `end_time` properties and new `RelationshipType` entries.

part of #426


meretp (approved):
Thanks, lgtm

PR #637, opened 2023-05-09
fix licensing profile

meretp:
- update import paths in __init__.py
- remove dataclass annotation for classes without fields


armintaenzertng (approved):
Thanks!

PR #638, opened 2023-05-09
[issue-426] add SoftwareArtifact

armintaenzertng:
part of #426

### DIFF
--- a/src/spdx_tools/spdx3/model/software/software_artifact.py
+++ b/src/spdx_tools/spdx3/model/software/software_artifact.py
@@ -1,0 +1,24 @@
+# SPDX-FileCopyrightText: 2023 spdx contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+from abc import abstractmethod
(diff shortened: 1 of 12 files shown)


src/spdx_tools/spdx3/model/software/software_artifact.py
```diff
@@ -0,0 +1,26 @@
+#  SPDX-FileCopyrightText: 2023 spdx contributors
+#
+#  SPDX-License-Identifier: Apache-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
```
meretp, line 5:
```suggestion
```

meretp (approved):
Thanks, jsut one minor remark from my side

src/spdx_tools/spdx3/model/software/software_artifact.py
armintaenzertng: That's some stupid PyCharm bug, thanks for catching!

Issue #432, opened 2023-01-18
implement fixtures for tests on the prototype spdx 3


meretp:
Analoguous to the fixtures in spdx 2.x we also want to use fixtures for the tests of the prototyped model.

HarshvMahawar:
is this issue still open? wanna work on it

armintaenzertng:
This targets the `prototype-spdx-3.0` branch. Are you familiar with the current working state of the SPDX 3.0 model? If yes, go ahead, I have unassigned myself from this issue.

HarshvMahawar:
fixtures as in the sense of having sample documents on which newly formed functions in "prototype model 3.0" can be tested?

armintaenzertng:
Yes, exactly. Basically an analogue of this file: https://github.com/spdx/tools-python/blob/refactor-python-tools/tests/spdx/fixtures.py

HarshvMahawar:
what specific program files do we have to target? and what difference it might have from the fixtures in spdx 2.x? from what I can see there are various class methods, do the fixture file have to be made around those?

armintaenzertng:
Ideally, we have a fixture for every class in this folder: https://github.com/spdx/tools-python/tree/prototype-spdx-3.0/src/spdx3/model.
The difference to the 2.x fixtures is that the classes and class structures have changed in 3.0, but the principle should be very much the same.

HarshvMahawar:
yaa I am working on it but it will take time as my schedule is tight right now

armintaenzertng:
No pressure, I just wanted to know if you still plan to work on it :)

armintaenzertng:
Hi again, any estimation on when you might put a PR up for this? We'll probably need the fixtures as input data for the upcoming serialization efforts.
If you need any help, please reach out!

HarshvMahawar:
Sorry for that, but for me this month was quite hectic due to various
examination. If I find any difficulty I will definitely reach out to you.
Thanks and Regards
HarshvM

HarshvMahawar:
@armintaenzertng What sample details should I put in the creation_info class for testing the classes, also are there any references from where I can know what are the valid sample strings or values that can be used for testing as there are some new variables that have been introduced?

armintaenzertng:
[...]

PR #641, opened 2023-05-10
[issue-432] rework SPDX3 Core fixtures

armintaenzertng:
This makes the existing Core fixtures more consistent, uses URIs where appropriate and fixes the handling of lists as standard arguments.

part of #432


meretp (approved):
Thanks, lgtm

PR #642, opened 2023-05-10
update implementation according to recent changes

meretp:
- update AIPackage according to restrictions to external properties
- update model_data_explainability and model_data_preprocessing to be lists
- rename buildStart and buildEnd in Build Class
- add supplied_by and standard in Artifact
- add profileIdentifier in CreationInfo and update the Enum class accordingly
- make property "to" optional in relationship
- update Dataset according to restrictions to external properties
- add property datasetType
- change type hint for supplied_by to str in security relationships


src/spdx_tools/spdx3/bump_from_spdx2/package.py
```diff
@@ -37,7 +37,9 @@ def bump_package(
     # package.supplier -> Relationship, suppliedBy?
     print_missing_conversion("package2.supplier", 0, "https://github.com/spdx/spdx-3-model/issues/113")
```
armintaenzertng, line 38: supplier can now be bumped, too

src/spdx_tools/spdx3/bump_from_spdx2/package.py
armintaenzertng, line 1: While we are at it:
License comments can now be bumped, too (there is a property `licenseComment` on `License`). According to the [description in the 2.3 spec](https://spdx.github.io/spdx-spec/v2.3/package-information/#716-comments-on-license-field), this should go to the `concludedLicense`.
This probably also applies to files and snippets.

armintaenzertng (commented):
Thanks, lgtm, but there are likely some bugs we will have to fix down the road, I imagine.
Two comments regarding bumping.

src/spdx_tools/spdx3/bump_from_spdx2/package.py
armintaenzertng: Good questions, I also suggest opening an issue!

tests/spdx3/model/test_tool.py
```diff
@@ -26,7 +28,8 @@ def test_invalid_initialization():
         Tool(12, 345)
 
     assert err.value.args[0] == [
-        'SetterError Tool: type of argument "spdx_id" must be str; got int instead: 12',
+        'SetterError Tool: type of argument "spdx_id" must be str; got int instead: ' "12",
```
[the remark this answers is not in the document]
meretp: deleted

src/spdx_tools/spdx3/bump_from_spdx2/package.py
meretp: fixed

armintaenzertng (approved):
Thanks!
Let's merge this and open a separate issue for the conversion of license comments.

PR #645, opened 2023-05-11
[issue-644] change type of extension property

### LINKED ISSUE #644: Change type of `extension` property
While writing tests, I'd like to check that every property of a class is set. This is not possible so long as the type of `extension` is `None` and thus only allows `None` as a value.

armintaenzertng:
This changes the type of `extension` to `Optional[str]`.

fixes #644


armintaenzertng:
I messed this up, opening a new one.

PR #651, opened 2023-05-12
[issue-649] fix TV-parser 

### LINKED ISSUE #649: TV-Parser raises IndexError if package has no SPDXID
When trying to parse a tag value which has a package without spdxid which precedes a file (so implying a CONTAINS relationship) an uncaught `IndexError` is raised.

meretp:
With this PR we log an error if the build of a contains relationship fails due to a non-parseable preceding package.

fixes #649

### DIFF
--- a/tests/spdx/parser/tagvalue/test_tag_value_parser.py
+++ b/tests/spdx/parser/tagvalue/test_tag_value_parser.py
@@ -54,6 +54,31 @@ def test_building_contains_relationship():
     ]
 
 
+def test_build_contains_relationship_with_error():
+    parser = Parser()
+    file_spdx_ids = ["SPDXRef-File-in-Package", "SPDXRef-Second-File-in-Package"]
+    document_str = "\n".join(
+        [
+            DOCUMENT_STR,
+            "PackageName: Package with two files",
+            "PackageDownloadLocation: https://download.com",
+            "FileName: File in package",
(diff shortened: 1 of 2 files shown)


tests/spdx/parser/tagvalue/test_tag_value_parser.py
```diff
@@ -54,6 +54,31 @@ def test_building_contains_relationship():
...
+            "FileName: File in package",
+            f"SPDXID: {file_spdx_ids[0]}",
+            "FileChecksum: SHA1: d6a770ba38583ed4bb4525bd96e50461655d2759",
+            "FileName: Second file in package",
+            f"SPDXID: {file_spdx_ids[1]}",
+            "FileChecksum: SHA1: d6a770ba38583ed4bb4525bd96e50461655d2759",
+        ]
+    )
+    with pytest.raises(SPDXParsingError) as err:
+        _ = parser.parse(document_str)
```
armintaenzertng, line 74:
```suggestion
        parser.parse(document_str)
```
Assignments are not mandatory.

armintaenzertng (approved):
Thanks for catching and fixing this bug! :)

tests/spdx/parser/tagvalue/test_tag_value_parser.py
meretp: fixed

armintaenzertng (approved):

PR #652, opened 2023-05-12
[issue-650] raise ParsingError if elements don't have an SPDXID

### LINKED ISSUE #650: RDF-parser raises no error if an element has no SPDXID
When trying to parse a file that contains an element (file, package, snippet) without `SPDXID` no error is raised. Instead the element gets a random ID as it is a blank node (e.g. "N5c02cc1dd7c345b7abeaef214e1fa3e8"). The validation catches this as it is no valid SPDXID but I think this should raise an error when parsing.

meretp:
fixes #650

### DIFF
--- a/src/spdx_tools/spdx/parser/rdf/graph_parsing_functions.py
+++ b/src/spdx_tools/spdx/parser/rdf/graph_parsing_functions.py
@@ -86,8 +86,8 @@ def parse_enum_value(enum_str: str, enum_class: Type[Enum], prefix: str) -> Enum
         raise SPDXParsingError([f"Invalid value for {enum_class}: {enum_str}"])
 
 
-def parse_spdx_id(resource: URIRef, doc_namespace: str, graph: Graph) -> Optional[str]:
-    if not resource:
+def parse_spdx_id(resource: Union[URIRef, BNode], doc_namespace: str, graph: Graph) -> Optional[str]:
+    if not resource or isinstance(resource, BNode):
         return None
     if resource.startswith(f"{doc_namespace}#"):
         return resource.fragment
--- a/src/spdx_tools/spdx/parser/rdf/package_parser.py
+++ b/src/spdx_tools/spdx/parser/rdf/package_parser.py
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: 2023 spdx contributors
 #
 # SPDX-License-Identifier: Apache-2.0
-from typing import Optional
+from typing import Optional, Union
 
 from rdflib import DOAP, RDFS, Graph, URIRef
 from rdflib.term import BNode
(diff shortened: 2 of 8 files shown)


tests/spdx/parser/rdf/test_file_parser.py
```diff
@@ -35,3 +37,18 @@ def test_parse_file():
...
+
+
+def test_parse_invalid_file():
+    graph = Graph().parse(os.path.join(os.path.dirname(__file__), "data/invalid_documents/file_without_spdx_ids.xml"))
+    file_node = graph.value(predicate=RDF.type, object=SPDX_NAMESPACE.File)
+    doc_namespace = "https://some.namespace"
+
+    assert isinstance(file_node, BNode)
+    with pytest.raises(SPDXParsingError) as err:
+        _ = parse_file(file_node, graph, doc_namespace)
```
armintaenzertng, line 49:
```suggestion
        parse_file(file_node, graph, doc_namespace)
```
Also at the other locations.

src/spdx_tools/spdx/parser/rdf/graph_parsing_functions.py
```diff
@@ -140,7 +140,7 @@ def get_value_from_graph(
     # this is a helper method to cast some rdf types from graph.value() to be compatible with the
     # code that follows
     value = graph.value(subject=subject, predicate=predicate, object=_object, default=default, any=_any)
-    if value and not isinstance(value, (URIRef, Literal, BNode)):
+    if value != _any and value and not isinstance(value, (URIRef, Literal, BNode)):
```
armintaenzertng, line 143: I don't get the `value != _any` part here. `_any` is a bool, why should it be equal to `value`?

src/spdx_tools/spdx/parser/rdf/package_parser.py
```diff
@@ -108,7 +108,9 @@ def parse_package(package_node: URIRef, graph: Graph, doc_namespace: str) -> Pac
     )
     homepage = parse_literal(logger, graph, package_node, DOAP.homepage)
     attribution_texts = []
-    for _, _, attribution_text_literal in graph.triples((package_node, SPDX_NAMESPACE.attributionText, None)):
+    for _, _, attribution_text_literal in get_correctly_typed_triples(
+        logger, graph, package_node, SPDX_NAMESPACE.attributionText, None
+    ):
```
armintaenzertng, line 113: Is this a separate issue? I don't see the connection to missing SPDX Ids

armintaenzertng (commented):
Thanks for the fix, I have a couple of questions, though.

src/spdx_tools/spdx/parser/rdf/graph_parsing_functions.py
meretp: Yeah, sorry, mixed the values here. This needs to be `value != default`. While working on this I noticed that an error about wrong types would be logged if a `package` has no `files_analyzed` field. This is as the changes you commented below not related to the origin issue and rather a fix for the wrapper methods introduced with #563. Shall I create a separate PR for these changes?

src/spdx_tools/spdx/parser/rdf/package_parser.py
meretp: Yes, as commented above a separate issue.

tests/spdx/parser/rdf/test_file_parser.py
meretp: fixed

src/spdx_tools/spdx/parser/rdf/graph_parsing_functions.py
armintaenzertng: Yes, please do :)

meretp: https://github.com/spdx/tools-python/pull/653 
The PR should be merged first as the test in this PR fails without these changes.

armintaenzertng (approved):
Thanks! :)

PR #658, opened 2023-05-16
[issue-427] fix external document ref bumping

armintaenzertng:
The new conversion follows the description provided in the [migration guide](https://docs.google.com/document/d/1-olHRnX1CssUS67Psv_sAq9Vd-pc81HF8MM0hA7M0hg/edit#heading=h.h8bkqwghh9fe).

The test is a little more verbose than necessary because I had some trouble wrapping my head around which field goes where.

Next up is adding external elements to the `imports` list, as also described in the migration guide.

part of #427

### DIFF
--- a/src/spdx_tools/spdx3/bump_from_spdx2/creation_information.py
+++ b/src/spdx_tools/spdx3/bump_from_spdx2/creation_information.py
@@ -23,10 +23,14 @@ def bump_creation_information(spdx2_creation_info: Spdx2_CreationInfo, payload: 
     print_missing_conversion("creation_info.document_namespace", 0, "https://github.com/spdx/spdx-3-model/issues/87")
 
     # creation_info.external_document_refs -> spdx_document.imports
-    imports = [
-        bump_external_document_ref(external_document_ref)
-        for external_document_ref in spdx2_creation_info.external_document_refs
-    ]
+    namespaces, imports = zip(
+        *[
+            bump_external_document_ref(external_document_ref)
+            for external_document_ref in spdx2_creation_info.external_document_refs
+        ]
+    )
+    namespaces = list(namespaces)
+    imports = list(imports)
     # creation_info.license_list_version -> ?
     print_missing_conversion(
         "creation_info.license_list_version",
@@ -76,4 +80,5 @@ def bump_creation_information(spdx2_creation_info: Spdx2_CreationInfo, payload: 
         element=[],
         root_element=[],
         imports=imports,
+        namespaces=namespaces,
     )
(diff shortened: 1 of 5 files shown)


src/spdx_tools/spdx3/bump_from_spdx2/creation_information.py
```diff
@@ -23,10 +23,14 @@ def bump_creation_information(spdx2_creation_info: Spdx2_CreationInfo, payload:
     print_missing_conversion("creation_info.document_namespace", 0, "https://github.com/spdx/spdx-3-model/issues/87")
 
     # creation_info.external_document_refs -> spdx_document.imports
```
meretp, line 25: Probably either delete this comment or update to "-> spdx_document.namespaces, spdx_document.imports".

In general I think we can delete these comments. In the beginning I added them to get a better overview on the migration but with the migration guide I don't think that we need them any longer.

src/spdx_tools/spdx3/bump_from_spdx2/creation_information.py
```diff
@@ -23,10 +23,14 @@ def bump_creation_information(spdx2_creation_info: Spdx2_CreationInfo, payload:
...
-        bump_external_document_ref(external_document_ref)
-        for external_document_ref in spdx2_creation_info.external_document_refs
-    ]
+    namespace_map, imports = zip(
+        *[
+            bump_external_document_ref(external_document_ref)
+            for external_document_ref in spdx2_creation_info.external_document_refs
+        ]
+    )
+    namespace_map = list(namespace_map)
```
meretp, line 32:
```suggestion
    namespaces = list(namespace_map)
```

tests/spdx3/bump/test_external_document_ref_bump.py
```diff
@@ -0,0 +1,36 @@
...
+from spdx_tools.spdx3.bump_from_spdx2.external_document_ref import bump_external_document_ref
+from spdx_tools.spdx3.payload import Payload
+from spdx_tools.spdx.model import ExternalDocumentRef
+from tests.spdx.fixtures import checksum_fixture, creation_info_fixture, external_document_ref_fixture
+
+
+def test_bump_external_document_ref():
+    checksum = checksum_fixture()
+    external_document_ref = ExternalDocumentRef("DocumentRef-external", "https://external.uri", checksum)
+    namespace_map, imports = bump_external_document_ref(external_document_ref)
```
meretp, line 16:
```suggestion
    namespace_map, external_map = bump_external_document_ref(external_document_ref)
```

meretp (approved):
Thanks, lgtm. I have two suggestions for renaming that I think will make the code more understandable but that's up to you.

tests/spdx3/bump/test_external_document_ref_bump.py
armintaenzertng: Well, the property is actually called `imports`, with its type being `ExternalMap`, so I probably leave it like that for now.

PR #660, opened 2023-05-16
fix tests after typeguard update

armintaenzertng:
The last typeguard update changed the error message text; to ensure this will not break again in the future, this will now not be tested anymore.

Some of the negative tests might look a little pointless right now because in a next step, I will change these to test all properties just like it is already done in [test_element_and_licensing_subclasses.py](https://github.com/spdx/tools-python/blob/6b56465a934fa2ba5b7d04f55e82f6971cb51671/tests/spdx3/model/test_element_and_licensing_subclasses.py).


armintaenzertng:
Yes, I have also thought about adapting the SPDX2 model tests as they are _very_ verbose right now.

PR #661, opened 2023-05-17
[issue-432] use fixtures in initialization tests

armintaenzertng:
part of #432

This also tests that every property is getting a value in the fixtures.


meretp (approved):
Thanks! I was at first confused why we should check all properties to not be `None` but this ensures that we add the property not only in the model but also in the fixture, I like that!

PR #665, opened 2023-05-17
[issue-659] make CreationInfo optional

### LINKED ISSUE #659: rename CreationInformation to CreationInfo
With https://github.com/spdx/spdx-3-model/pull/175 the class `CreationInformation` was renamed to `CreationInfo`, to align with the specification we should also rename the class [CreationInformation](https://github.com/spdx/tools-python/blob/prototype-spdx-3.0/src/spdx_tools/spdx3/model/creation_information.py).

armintaenzertng:
fixes #659

I also fixed some inconsistencies in the order of relationship properties.


meretp (approved):
Thanks! As the `CreationInfo` is optional now, there is no need to include it (resp. the mock) in the tests any longer. But since the tests will be refactored soon (and probably also the decision to make this fiel optional), I think it's not worth to delete it now.

PR #667, opened 2023-05-17
Add fixture dictionaries for software profile

HarshvMahawar:
targets #432

### DIFF
--- a/tests/spdx3/fixtures.py
+++ b/tests/spdx3/fixtures.py
@@ -36,6 +36,7 @@ from spdx_tools.spdx3.model import (
 from spdx_tools.spdx3.model.licensing import (
     CustomLicense,
     CustomLicenseAddition,
+    LicenseField,
     ListedLicense,
     ListedLicenseException,
 )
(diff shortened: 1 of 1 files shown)


tests/spdx3/fixtures.py
```diff
@@ -298,6 +309,42 @@ def namespace_map_fixture(
...
+    "originated_by": ["https://spdx.test/tools-python/originatedBy"],
+    "supplied_by": ["https://spdx.test/tools-python/suppliedBy"],
+    "built_time": datetime(2004, 1, 1),
+    "release_time": datetime(2005, 1, 1),
+    "valid_until_time": datetime(2006, 1, 1),
+    "standard": ["https://spdx.test/tools-python/standard"],
+}
+
+SOFTWARE_ARTIFACT_DICT = {
+    "content_identifier": "contentIidentifier",
```
armintaenzertng, line 322: `content_identifier` must be a URI, see [here](https://github.com/spdx/spdx-3-model/blob/main/model/Software/Properties/contentIdentifier.md)
```suggestion
    "content_identifier": ""https://spdx.test/tools-python/contentIdentifier"",
```

tests/spdx3/fixtures.py
```diff
@@ -298,6 +309,42 @@ def namespace_map_fixture(
...
+    "purpose": SoftwarePurpose.OTHER,
+    "concluded_license": LicenseField,
+    "declared_license": LicenseField,
```
armintaenzertng, line 325: This is not a correct value for these fields. But I suppose you ran into the circle definition of the `License` fixture being referenced in the same dictionary as the software artifact fixtures. I'll think about how to fix this, in the meantime we can merge it like this, I think.

armintaenzertng (changes requested):
Thanks, this looks very good already, I have only one minor remark.

tests/spdx3/fixtures.py
HarshvMahawar: yes

HarshvMahawar: Also, do the values in SNIPPET_DICT assumed correctly, or need modifications?

armintaenzertng: They are fine as far as I'm concerned! :)

armintaenzertng (approved):
Looks good to me, thanks!

Issue #673, opened 2023-05-26
More performant type checking


armintaenzertng:
Parsing of large SPDX files takes a long time with the current use of `typeguard` to check for correct types during initialization. For reference: The spdx file linked in #672 takes 190 seconds to parse.
It might be possible that other libraries like `pydantic` are more performant; but maybe we will have to come up with an additional validation layer that can be disabled for very large documents.

armintaenzertng:
Update: #685 introduces `beartype` instead of `typeguard`.

PR #675, opened 2023-05-26
[issue-673] remove type checking due to performance issues

armintaenzertng:
This is a quick fix for #673 that replaces the custom `@dataclass_with_properties` with the classic `@dataclass` for a massive performance gain. (From 190s down to 3s for a large spdx example file).

There will need to be some clean-up in a follow-up PR; for example the `__init__` are not needed anymore now.


fholger (dismissed):

maxhbr (dismissed):
changing `dataclass_with_properties` to `dataclass` causes a lot of changes in many files and if we see that we still need this wrapper, it is a huge change to roll back.

maxhbr (approved):
LGTM

PR #676, opened 2023-05-26
Draft: Fix shacl validation

fholger:
Make the test json ld file successfully verify against the shacl model file.

### DIFF
--- a/src/spdx_tools/spdx3/writer/json_ld/context.json
+++ b/src/spdx_tools/spdx3/writer/json_ld/context.json
@@ -1,5 +1,14 @@
 {
+    "ai": "https://spdx.org/rdf/AI/",
+    "build": "https://spdx.org/rdf/Build/",
     "core": "https://spdx.org/rdf/Core/",
+    "dataset": "https://spdx.org/rdf/Dataset/",
+    "licensing": "https://spdx.org/rdf/Licensing/",
+    "ns0": "http://www.w3.org/2003/06/sw-vocab-status/ns#",
+    "owl": "http://www.w3.org/2002/07/owl#",
+    "rdfs": "http://www.w3.org/2000/01/rdf-schema#",
+    "security": "https://spdx.org/rdf/Security/",
+    "sh": "http://www.w3.org/ns/shacl#",
     "software": "https://spdx.org/rdf/Software/",
     "xsd": "http://www.w3.org/2001/XMLSchema#",
     "AIPackage": "ai:AIPackage",
@@ -8,7 +17,12 @@
     "AnonymousPayload": "core:AnonymousPayload",
     "Organization": "core:Organization",
     "Person": "core:Person",
+    "SoftwareAgent": "core:SoftwareAgent",
     "SpdxDocument": "core:SpdxDocument",
+    "spdxId": {
+        "@id": "core:spdxId",
+        "@type": "xsd:anyURI"
+    },
     "Dataset": "dataset:Dataset",
     "ConjunctiveLicenseSet": "licensing:ConjunctiveLicenseSet",
     "CustomLicense": "licensing:CustomLicense",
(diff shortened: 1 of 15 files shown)


src/spdx_tools/spdx3/writer/json_ld/context.json
```diff
@@ -1,14 +1,28 @@
...
     "Annotation": "core:Annotation",
     "AnonymousPayload": "core:AnonymousPayload",
     "Organization": "core:Organization",
     "Person": "core:Person",
+    "SoftwareAgent": "core:SoftwareAgent",
     "SpdxDocument": "core:SpdxDocument",
+    "spdxId": {
+        "@id": "core:spdxId",
+        "@type": "xsd:anyURI"
+    },
```
armintaenzertng, line 25: `spdxId` is `@id` in json-ld

src/spdx_tools/spdx3/writer/json_ld/context.json
```diff
@@ -183,7 +193,7 @@
     },
     "createdBy": {
         "@id": "core:createdBy",
-        "@type": "core:Entity"
+        "@type": "@id"
```
armintaenzertng, line 196: To my knowledge, the type is still `core:Agent`. If there is only a string/URI present then it will be taken as a reference to the object with the corresponding URI, which should then be of type `core:Agent`.
(also applies to all other occurrences like this)

src/spdx_tools/spdx3/writer/json_ld/context.json
```diff
@@ -264,7 +278,7 @@
         "@type": "xsd:anyURI"
     },
     "locator": {
-        "@id": "core:locator",
+        "@id": "security:locator",
```
armintaenzertng, line 281: There are two properties named `locator` in the spec, one in `core`, one in `security`. We can expand either one of them (in that case I would go with `core`) or neither (i.e. keep using `core:locator`and `security:locator` everywhere).

src/spdx_tools/spdx3/writer/json_ld/context.json
```diff
@@ -330,19 +348,19 @@
...
+        "@type": "@id"
     },
     "summary": {
         "@id": "core:summary",
         "@type": "xsd:string"
     },
     "suppliedBy": {
-        "@id": "core:suppliedBy",
-        "@type": "core:Agent"
+        "@id": "security:suppliedBy",
```
armintaenzertng, line 358: Same thing as with `locator`.

armintaenzertng (commented):
Some comments on the `context.json`.

src/spdx_tools/spdx3/writer/json_ld/context.json
armintaenzertng: That is, `spdxId` will probably never pop up in json-ld

src/spdx_tools/spdx3/writer/json_ld/context.json
fholger: That would have been great, but unfortunately not what it did. Instead, it interpreted the URI as a literal and then complained that it did not have the expected type. Only by setting this to `@id` it works as expected. Shacl will still validate that the entity pointed to has the expected type due to the `sh:class` constraint.

src/spdx_tools/spdx3/writer/json_ld/context.json
fholger: It was still included in the model as a (dead) reference, should be gone now.

src/spdx_tools/spdx3/writer/json_ld/context.json
fholger: I've added a check to prefer and keep core properties here.

src/spdx_tools/spdx3/writer/json_ld/json_ld_converter.py
```diff
@@ -69,7 +69,7 @@ def _convert_to_json_ld_dict(element: Any, alt_creation_info=False, alt_hash=Fal
         elif attribute_value:
             if attribute_name == "_spdx_id":
                 attribute_name = "@id"
-            if attribute_name == "_from_element":
+            elif attribute_name == "_from_element":
```
armintaenzertng, line 72: The underscores in front of the property names originate in our `dataclass_with_properties` implementation, which is currently under rework. Just saying that we possibly need to adapt this once we rebase.

PR #678, opened 2023-05-30
[issue-677] point installation instructions to newest release

### LINKED ISSUE #677: ModuleNotFoundError: No module named 'spdx_tools'
Can't seem to install and reference spdx-tools.
We've tried in a virtual venv and just straight up python3 environment like in the example.

` pip3 install spdx-tools`

but when we try to reference it we invariably get something like:

```
    from spdx_tools.spdx.model import (Checksum, ChecksumAlgorithm, File,
ModuleNotFoundError: No module named 'spdx_tools'

```
could there be a step we're missing?

armintaenzertng:
fixes #677

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,8 +56,9 @@ The implementation is based on the descriptive markdown files in the repository 
 # Installation
 
 As always you should work in a virtualenv (venv). You can install a local clone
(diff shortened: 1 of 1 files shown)


maxhbr (approved):
LGTM

PR #685, opened 2023-06-01
Replace typeguard with beartype

fholger:
As typeguard was found to be too slow for larger input files, this PR replaces it with beartype. beartype is significantly faster and can handle large inputs, however, it comes with a caveat in that it does not do full nested deep type checks, but instead does random sampling in such instances. As such, it can allow certain misuses of types to go undetected, but it will still detect a majority of typical type mistakes. It is most likely better to have beartype than have nothing now even if it ends up not being the final solution to the validation issue.


tests/spdx/parser/jsonlikedict/test_error_message.py
armintaenzertng, line 1: We might want to keep this kind of test so that we retain some control over what the error messages would actually look like. Point in case: The latest typeguard update changed the error message for union types to an incomprehensible multi-liner.

armintaenzertng (approved):

tests/spdx/parser/jsonlikedict/test_error_message.py
fholger: Unfortunately, beartype error messages are a bit randomized in the order of union types, so this test would need to be done a bit differently. I propose dropping it for now, and if we stick with beartype longer term, we can have a more detailed look into its error messages and add appropriate tests and improvements to them.

PR #686, opened 2023-06-01
fix dataset sensor property

### LINKED ISSUE #682: [Bug] Sensor is required for Dataset object
If I try to create a `Dataset` object using the `prototype-spdx-3.0` branch, and found that when you try to create one like so:

```python
ds = Dataset(
    spdx_id=dataset_id,
    creation_info=creation_info,
    name=dataset.name,
    dataset_type="text",
    originated_by=["s3"],
    download_location=os.getcwd(),
    purpose=[SoftwarePurpose.DATA],
    built_time=datetime.now(),
    release_time=datetime.now(),
)
```

It fails with the following error:

```
Traceback (most recent call last):
  File "/Users/williamarmiros/protectai-code/bom/spdx.py", line 110, in __init__
    ds = Dataset(
  File "/Users/williamarmiros/open-source/spdx/tools-python/src/spdx_tools/spdx3/model/dataset/dataset.py", line 100, in __init__
  File "/Users/williamarmiros/open-source/spdx/tools-python/src/spdx_tools/common/typing/type_checks.py", line 37, in check_types_and_set_values
    raise ConstructorTypeErrors(errors)
spdx_tools.common.typing.constructor_type_errors.ConstructorTypeErrors: ['SetterError Dataset: argument "sensor" (None) is not a dict: None']
```

The problem is the `sensor` property cannot be a `None` type, which should be allowed by the spec. I will submit a PR to fix this tomorrow!

willarmiros:
Fixes #682

### DIFF
--- a/src/spdx_tools/spdx3/model/dataset/dataset.py
+++ b/src/spdx_tools/spdx3/model/dataset/dataset.py
@@ -94,7 +94,7 @@ class Dataset(Package):
         supplied_by = [] if supplied_by is None else supplied_by
         standard = [] if standard is None else standard
         data_preprocessing = [] if data_preprocessing is None else data_preprocessing
-        sensors = {} if sensor is None else sensor
+        sensor = {} if sensor is None else sensor
         known_bias = [] if known_bias is None else known_bias
         anonymization_method_used = [] if anonymization_method_used is None else anonymization_method_used
         check_types_and_set_values(self, locals())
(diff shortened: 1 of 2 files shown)


armintaenzertng (approved):
Thanks! :)

willarmiros:
Added one more fix for the dataset debug writer :)

willarmiros:
@armintaenzertng any chance for a merge here? :)

PR #696, opened 2023-06-14
POC: JSON-LD parsing without rdflib

armintaenzertng:
Important feature is the module with the string to class mappings, which we will probably need for every serialization format.

Also adds a full workflow test, consisting of conversion->writing->parsing->writing, then comparing the two written outputs.

### DIFF
--- a/src/spdx_tools/spdx3/parser/json/json_parser.py
+++ b/src/spdx_tools/spdx3/parser/json/json_parser.py
@@ -1,0 +1,114 @@
+# SPDX-FileCopyrightText: 2022 spdx contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+import json
+from typing import Union
+
+from beartype.typing import Dict, List
+from semantic_version import Version
+
+from spdx_tools.spdx3.model import HashAlgorithm
+from spdx_tools.spdx3.parser.model_definition_dicts import (
+    CLASS_DICT,
+    DATETIME_PROPERTIES,
+    ENUM_DICT,
+    KEYS_WITH_URI_VALUES,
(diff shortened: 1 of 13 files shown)


src/spdx_tools/spdx3/parser/json/json_parser.py
```diff
@@ -0,0 +1,114 @@
...
+
+def parse_list_of_elements(list_of_elements: List[Dict], custom_context: Dict):
+    payload = Payload()
+    for element_object in list_of_elements:
+        payload.add_element(parse_element_object(element_object, custom_context))
+    return payload
+
+
+def parse_element_object(element_object: Union[Dict, str], custom_context: Dict):
+    if isinstance(element_object, str):
```
fholger, line 53: In general, JSON also supports numbers, booleans and null directly. Don't we need to handle them in similar fashion?

src/spdx_tools/spdx3/parser/json/json_parser.py
```diff
@@ -0,0 +1,114 @@
...
+                value = [expand_uri(entry, custom_context) for entry in value]
+            else:
+                value = expand_uri(value, custom_context)
+
+        if key in SPECIALLY_TREATED_KEYS:
+            element_properties[SPECIALLY_TREATED_KEYS[key]] = value
+        else:
+            element_properties[camel_case_to_snake_case(key)] = value
+
+    if element_type not in CLASS_DICT:
```
fholger, line 85: This could be moved before the `for` loop?

maxhbr:
This is stale, and https://github.com/JPEWdev/shacl2code is expected to replace it

PR #698, opened 2023-06-15
Update spdx2 to 3 conversion

armintaenzertng:
This fixes a few things that came up during writing the parser:
- use `rootElement` for DESCRIBED Elements
- add the correct URI prefix before the data license
- don't write `CreationInfo` on all Elements (may have to change in the future depending on how we will treat this)
- don't convert licenses (licensing format has yet to be decided on)


maxhbr (approved):
LGTM

PR #700, opened 2023-06-20
Expand writer API to also allow serialization to streams / file objects

fholger:
As of now, our writers always expect a file name to be provided and then open an output file to write to the given name. This PR adds a set of function that instead allow to pass in a stream / file object, instead. This allows us to pass other types of streams (e.g. in-memory streams) rather than just files and thus makes the API more flexible.


armintaenzertng (approved):
Looks good to me, thanks for the addition :)

armintaenzertng (approved):
Thanks for the hard work! :)

PR #702, opened 2023-06-21
Create new Github workflow to build a release when a tag is pushed

fholger:
Whenever a versioned tag (vX.Y.Z) is pushed, this workflow will build the python wheel, create a new GitHub release for the pushed tag and attach the build output to the release.
This is a first step towards automating our release process. In the future, we may want to directly upload the release to PyPI.

Sample run: https://github.com/fholger/tools-python/actions/runs/5332473607/jobs/9661695249
Sample created release: https://github.com/fholger/tools-python/releases/tag/v0.8.1a4

### DIFF
--- a/.github/workflows/prepare_release.yml
+++ b/.github/workflows/prepare_release.yml
@@ -1,3 +1,7 @@
+# SPDX-FileCopyrightText: 2023 spdx contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+
 name: Prepare release
 
 on:
--- a/.github/workflows/prepare_release.yml
+++ b/.github/workflows/prepare_release.yml
@@ -1,0 +1,40 @@
+name: Prepare release
+
+on:
+  push:
+    tags: [ 'v*.*.*']
+
+jobs:
+  release:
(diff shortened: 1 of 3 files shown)


.github/workflows/prepare_release.yml
```diff
@@ -0,0 +1,39 @@
...
+        run: python -m build
+      - name: Verify build
+        run: twine check dist/*
+      - name: Create build archive
+        uses: a7ul/tar-action@v1.1.0
+        with:
+          command: c
+          files: dist
+          outPath: spdx_tools_dist.tar.gz
+      - name: Create GitHub release
```
maxhbr, line 35: Can you add functionality to just create a prerelease instead of a full release for alpha, beta and rc

maxhbr (approved):
Just a copyright header is missing. Looks good to me

.github/workflows/prepare_release.yml
fholger: As discussed, the releases are now created as drafts for simplicity, so that someone can look them over and mark them appropriately before publishing them.

Issue #689, opened 2023-06-07
Automatic checks for missing license headers


armintaenzertng:
There is currently no tool in the CI that checks each file for a correct license header. We should introduce such a check, for example by using `reuse`.

goneall:
Just for reference, this was an Google Summer of Code project to perform a similar function.  It is no longer being maintained and is archived here: https://github.com/spdx/license-coverage-grader

harshkasat:
Hey there! I am a beginner. Can you guide me on how to resolve this issue? Do we have to run an auto-check test?

PR #703, opened 2023-06-23
Add missing license headers in source files

fholger:
Inspired by #689 I ran the reuse tool on the code base. It reported a number of missing license and/or copyright headers in our Python source files, which this PR addresses.

Note that reuse also complains about a number of auxiliary files, so this PR does not achieve full compliance on its own.


maxhbr (approved):
LGTM

PR #708, opened 2023-06-26
Create tool model files from spdx-3-model files

davaya:
Adds two scripts:
1) `load_model.py` parses model files from the spdx-3-model repo and saves the full model as a single file modelTypes.json.  This is complete and may be generally useful.
2) 'make-model-classes.py' reads modelTypes.json and creates class files from it.  This is currently a skeleton that generates only boilerplate class files, to illustrate what needs to be developed.

### DIFF
--- a/src/spdx_tools/spdx3/load_model.py
+++ b/src/spdx_tools/spdx3/load_model.py
@@ -1,0 +1,119 @@
+import json
+import os
+import re
+from io import TextIOWrapper
+from typing import TextIO
+from urllib.request import urlopen, Request
+from urllib.parse import urlparse
+
+"""
+Read SPDX model files from GitHub or a local copy (SPDX_MODEL), parse and save into a single file (GENERATED_MODEL)
+"""
+
+SPDX_MODEL = 'https://api.github.com/repos/spdx/spdx-3-model/contents/model'
+GENERATED_MODEL = 'model-generated'
+
+AUTH = {'Authorization': f'token {os.environ["GitHubToken"]}'}  # Prevent rate limiting
+
+
+class WebDirEntry:
+    """
+    Fake os.DirEntry type for GitHub filesystem
+    """
+    def __init__(self, name, path, url):
+        self.name = name
+        self.path = path
+        self.url = url
+
+
+def list_dir(dirpath: str) -> dict:
+    """
+    Return a dict listing the files and directories in a directory on local filesystem or GitHub repo.
+
+    :param dirpath: str - a filesystem path or GitHub API URL
+    :return: dict {files: [DirEntry*], dirs: [DirEntry*]}
+    Local Filesystem: Each list item is an os.DirEntry structure containing name and path attributes
+    GitHub Filesystem: Each list item has name, path, and url (download URL) attributes
+    """
+
+    files, dirs = [], []
+    u = urlparse(dirpath)
+    if all([u.scheme, u.netloc]):
+        with urlopen(Request(dirpath, headers=AUTH)) as d:
+            for dl in json.loads(d.read().decode()):
+                url = 'url' if dl['type'] == 'dir' else 'download_url'
+                entry = WebDirEntry(dl['name'], dl[url], dl['url'])
+                (dirs if dl['type'] == 'dir' else files).append(entry)
+    else:
+        with os.scandir(dirpath) as dlist:
+            for entry in dlist:
+                (dirs if os.path.isdir(entry) else files).append(entry)
+    return {'files': files, 'dirs': dirs}
+
+
+def open_file(fileentry: os.DirEntry) -> TextIO:
+    u = urlparse(fileentry.path)
+    if all([u.scheme, u.netloc]):
+        return TextIOWrapper(urlopen(Request(fileentry.path, headers=AUTH)), encoding='utf8')
+    return open(fileentry.path, 'r', encoding='utf8')
+
+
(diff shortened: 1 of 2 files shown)


goneall:
Thanks @davaya - Can you attach a sample output `modelTypes.json` file?  I may attempt something similar for the Java tools and would like to see an example.

davaya:
@goneall it's a straight translation of the markdown to JSON, plus four derived metadata properties prefixed with underscore, to make it easier to process.

https://gist.github.com/davaya/9f931c4b4cb36aa9dead21b4ee45b4a5

armintaenzertng:
Thanks for the work, David! However, I don't think we should keep a parser of the spec md-files in the python-tools repo. This is likely to go out of sync eventually. The spec-parser repo already takes care of this kind of parsing and should be the go-to for such endeavors, I think.
There [is also a PR](https://github.com/spdx/spec-parser/pull/54) ready to be merged that generates a JSON file similar to yours from the model. Please have look if that would satisfy your expectations! :)

davaya:
It did feel odd to have code where I put it since there was no previous example - I just wanted to make the parser available somewhere since it can read directly from the GitHub repo, it does zero manipulation of the model data (it just checks that the directory structure is as expected, then sucks in whatever it finds), and the class generator needs something concrete to work from.  Plus it's short - 120 lines total, 30 of which are the GitHub virtual filesystem. 

I'll move this to the spec-parser repo, though having a whole repo for one simple task seems like overkill.  @zvr 's md format is brilliant - simple enough even a caveman can parse it :-)

armintaenzertng:
We opened an issue on the auto-generation of the model from the spec: #716.
@davaya: Are you currently working on this topic? If not, we may take this on.

davaya:
@armintaenzertng I am currently working on it in the [Playground](https://github.com/spdx/spdx-3-serialization-prototype-playground/tree/main/model), and am glad to see #716 - that's a great step forward.  Please copy or ignore any of the playground code as you see fit.

davaya:
@armintaenzertng I am currently working on it in the [Playground](https://github.com/spdx/spdx-3-serialization-prototype-playground/tree/main/model), and am glad to see #716 - that's a great step forward.  Please copy or ignore any of the playground code as you see fit.

I'm tired of arguing about whether the spec-parser repo is the "one and only" thing allowed to read model files, that isn't the open source way.  I happen to think my [parser](https://github.com/spdx/spdx-3-serialization-prototype-playground/blob/main/model/parse_model.py) is much simpler and thus easier to verify for correctness than the spec-parser repo, but the code is only a suggestion, nobody is forced to use it or even look at it.

If you use some of its ideas and also manage to get them merged into spec-parser, that would be a wonderful thing.

davaya:
@armintaenzertng, @goneall  I am currently working on it in the [Playground](https://github.com/spdx/spdx-3-serialization-prototype-playground/tree/main/model), and am very happy to see #716 and #738 - that's a great step forward.  Please copy or ignore any of the playground code as you see fit.

I'm tired of arguing about whether the spec-parser repo is the one and only thing allowed to read model files, that isn't the open source way.  I happen to think my [parser](https://github.com/spdx/spdx-3-serialization-prototype-playground/blob/main/model/parse_model.py) is much simpler and thus much easier to verify for correctness than the spec-parser repo, but the code is only a suggestion, nobody is forced to use it or even look at it.

I don't see "External properties restrictions" as adding any value over putting the restrictions directly into the property being restricted, but since they were added to the model files, the playground class generator now supports doing restrictions both ways.

The playground class generator pre-processes class inheritance to ensure that types explicitly show all the properties that they can have.  That representation means it can be implemented in any language that doesn't support classes.  But I'm not sure any languages like that even exist any more, so Python and all other tools languages would presumably implement the model using OOP classes that do inherit properties.

If you use some of its ideas and also manage to get them merged into spec-parser, that would be a wonderful thing.  The JSON snapshot produced by the playground parser omits nothing from the model files (except summary and description content, which could easily be put back in), so the spec-parser repo code could get everything that exists in the markdown files from a JSON snapshot file.  But I'm getting a "not invented here" vibe against even looking at PRs for the spec-parser, much less considering technical merit, completeness or correctness.  Maybe you'll have better luck.

goneall:
> I'm tired of arguing about whether the spec-parser repo is the one and only thing allowed to read model files, that isn't the open source way. I happen to think my [parser](https://github.com/spdx/spdx-3-serialization-prototype-playground/blob/main/model/parse_model.py) is much simpler and thus much easier to verify for correctness than the spec-parser repo, but the code is only a suggestion, nobody is forced to use it or even look at it.

My only concern with supporting multiple parsers from the markdown files is it will require "standardizing" the markdown and may restrict our ability to quickly change the files.  The reason some months ago we chose to standardize on the OWL / SHACL rather than the markdown was the OWL / SHACL is already a well established stable spec.

If we're willing to allow rapid change of the markdowns and not require standardizing either the markdown nor the intermediate json files, I'm flexible.

davaya:
The playground parser doesn't know anything about the SPDX model, the only thing it knows are filesystem directories and markdown section headers, level 1 list items and level 2 list items.  As long as those things are "standard", the playground parser will read the entire `model` file hierarchy and produce a JSON snapshot of it.

If the markdown files started using some new markdown feature like level 3 list items, or changed the filesystem hierarchy to be more than 3 layers deep (currently model/profile/category/file.md) then the parser would need to be tweaked.

PR #723, opened 2023-07-05
[issue-722] add `calculate_package_verification_code()`

### LINKED ISSUE #722: Add back package verification code calculation
The functions to calculate file checksums and package verification codes have been removed during the refactoring for v0.8.
However, I found out that they are indeed being used, [for example by scancode](https://github.com/nexB/scancode-toolkit/blob/a15174f31efaf8816e8c9a65c9f85c4beffc0227/src/formattedcode/output_spdx.py#L406).
To allow libraries to upgrade, we should add back this functionality.

armintaenzertng:
also add `calculate_file_checksum()`, as this is needed by the package verification code

fixes #722

### DIFF
--- a/src/spdx_tools/spdx/spdx_element_utils.py
+++ b/src/spdx_tools/spdx/spdx_element_utils.py
@@ -1,9 +1,11 @@
 # SPDX-FileCopyrightText: 2022 spdx contributors
 #
 # SPDX-License-Identifier: Apache-2.0
+import hashlib
+
 from beartype.typing import List, Union
 
-from spdx_tools.spdx.model import ExternalDocumentRef, File, Package, Snippet
+from spdx_tools.spdx.model import ChecksumAlgorithm, ExternalDocumentRef, File, Package, Snippet
 
 
 def get_full_element_spdx_id(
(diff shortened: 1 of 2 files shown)


armintaenzertng:
It seems that Windows yields different results when using `calculate_file_checksum()` than Ubuntu does.

tests/spdx/test_checksum_calculation.py
```diff
@@ -0,0 +1,75 @@
...
+    unknown_file_name = "./unknown_file_name"
+    file1 = File(
+        unknown_file_name,
+        "SPDXRef-unknown",
+        [Checksum(ChecksumAlgorithm.MD4, "20862a6d08391d07d09344029533ec644fac6b21")],
+    )
+
+    with pytest.raises(FileNotFoundError) as err:
+        calculate_package_verification_code([file1])
+        assert unknown_file_name in err.value
```
fholger, line 75: This line is never reached, you need to unindent it one level.
You also have to explicitly convert the error to string, otherwise this assertion fails, i.e. `str(err.value)`

fholger (changes requested):

src/spdx_tools/spdx/spdx_element_utils.py
```diff
@@ -29,3 +31,41 @@ def get_full_element_spdx_id(
         raise ValueError(f"external id {external_id} not found in external document references")
 
     return external_uri + "#" + local_id
+
+
+def calculate_package_verification_code(files: List[File]) -> str:
+    list_of_file_hashes = []
+    hash_algorithm_name = ChecksumAlgorithm.SHA1
```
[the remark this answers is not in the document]
armintaenzertng: fixed

tests/spdx/test_checksum_calculation.py
armintaenzertng: Thanks! :)

fholger (approved):

PR #724, opened 2023-07-05
[issue-721] update Actor regex and parsing

### LINKED ISSUE #721: Is this organization name really invalid?
I'm getting the following error because of this Organization name:

~~~text
Organization: (c) Chris Sainty
~~~

The regex is the following:

~~~text
Organization:\s*(([^(])+)(\((.*)\))?
~~~

The documentation says the following:

~~~text
Organization: organization name and optional (<email>)
~~~

https://spdx.github.io/spdx-spec/v2.2.2/package-information/#76-package-originator-field

Should we update the regex inside this library, or is the organization name invalid?

armintaenzertng:
The new logic follows this rule:
First, split off the `ActorType` part.
Then, if the last character in the string is a closing bracket and there exists a corresponding opening bracket, match everything between these two brackets as `email` and the rest of the string as `name`.
If the the last character is not a closing bracket, match the whole string as `name`.

I also added some more tests for this.

fixes #721

### DIFF
--- a/src/spdx_tools/spdx/parser/actor_parser.py
+++ b/src/spdx_tools/spdx/parser/actor_parser.py
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 import re
 
-from beartype.typing import Match, Optional, Pattern
+from beartype.typing import Match, Pattern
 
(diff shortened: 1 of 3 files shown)


chrisdecker1201:
I think #721 is good. Thank you

PR #727, opened 2023-07-12
SPDX3: rename ProfileIdentifier to ProfileIdentifierType

fholger:
... to be consistent with the naming in the spec. This makes it easier to generate code working with our model as the naming is deterministic.


armintaenzertng (approved):
Assuming you used your IDE to catch all occurrences, this looks good to me ;)

PR #728, opened 2023-07-12
change return type of calculate_package_verification_code() to PackageVerificationCode

armintaenzertng:


fholger (approved):

HarshvMahawar:
what algorithm should the file hashes belong to? (like sha1, sha256 etc). Also, if the package contains sub-directory the hash should be of sub-directory right? thanks :)

PR #730, opened 2023-07-14
add script to publish from tag

maxhbr:
This probably would deserve a better placement, description and/or name

### DIFF
--- a/dev/publish_from_tag.sh
+++ b/dev/publish_from_tag.sh
@@ -1,0 +1,62 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: 2023 spdx contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+set -euo pipefail
+
+if [ $# -eq 0 ]; then
+    cat<<EOF
+Usage:
+  $0 \$VERSION
+Example:
+  $0 0.8.0rc2
+where 0.8.0rc2 is the version and v0.8.0rc2 is the corresponding tag.
+EOF
+    exit 0
+fi
+
+
+version="$1"
+tag="v${version}"
+tar_gz="https://github.com/spdx/tools-python/releases/download/${tag}/spdx_tools_dist.tar.gz"
+tag_dir="dist-${tag}"
+
+cd -- "$(dirname "$0")/.."
+
+if ! command -v twine &> /dev/null; then
+    echo "twine could not be found"
+    echo "maybe load venv with"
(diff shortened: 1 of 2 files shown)


publish_from_tag.sh
```diff
@@ -0,0 +1,50 @@
...
+
+cd -- "$(dirname "$0")"
+
+if ! command -v twine &> /dev/null; then
+    echo "twine could not be found"
+    echo "maybe load venv with"
+    echo "  . ./venv/bin/activate"
+    echo "  . ./venv/bin/activate.fish"
+    echo
+    echo "will try to avtivate..."
```
armintaenzertng, line 30:
```suggestion
    echo "will try to activate..."
```

publish_from_tag.sh
```diff
@@ -0,0 +1,55 @@
...
+tag_dir="dist-${tag}"
+
+cd -- "$(dirname "$0")"
+
+if ! command -v twine &> /dev/null; then
+    echo "twine could not be found"
+    echo "maybe load venv with"
+    echo "  . ./venv/bin/activate"
+    echo "  . ./venv/bin/activate.fish"
+    echo
```
armintaenzertng, line 31: maybe exit somewhere here? Else it will fail on line 49

armintaenzertng (commented):
In #726, we add a `dev` directory on top-level, this script might be well placed there.

dev/publish_from_tag.sh
```diff
@@ -0,0 +1,57 @@
...
+
+if [[ -d "$tag_dir" ]]; then
+    echo "the dir \"$tag_dir\" already exists, exiting for safety"
+    exit 1
+fi
+
+mkdir -p "$tag_dir"
+(cd "$tag_dir" && wget -c "$tar_gz" -O - | tar --strip-components=1 -xz)
+
+twine check "${tag_dir}/spdx-tools-${version}.tar.gz"  "${tag_dir}/spdx_tools-${version}-py3-none-any.whl"
```
armintaenzertng, line 51: This point is still reached even when `twine` is not installed.

armintaenzertng (dismissed):
My point from before still stands, but I wouldn't consider this blocking. Your choice to merge :)

dev/publish_from_tag.sh
maxhbr: now fixed

armintaenzertng (approved):

PR #738, opened 2023-07-25
Automatically generate SPDX3 model classes from spec

fholger:
An attempt to automatically generate the SPDX3 model classes from the spec to make it easier to keep the model representation up to date with the spec.

At this point it mostly does what it needs to do, but there are a few exceptions and special cases that still need to be addressed manually.


maxhbr:
This is stale, and https://github.com/JPEWdev/shacl2code is expected to replace it

PR #740, opened 2023-07-26
[issue-739] change order of properties when converting to JSON-dict

### LINKED ISSUE #739: Change order of properties when converting to json-dict
As of python 3.7, we are able to determine the order of dictionary entries by the order of insertion.
This will reflect in the order of the keys in the resulting JSON document.

We should use this to sort the most important keys (SPDXID, name, etc) to the top.

armintaenzertng:
fixes #739


armintaenzertng:
Closing due to controversy regarding the approach.

Issue #762, opened 2023-09-14
Tool is failing to parse URL like git@github.com:siimon/prom-client.git in downloadLocation


pushkargr:
URL like git@github.com:siimon/prom-client.git are valid download locations. Github provides the link this way. But the tool is marking this as invalid.

`package download_location must be a valid download location according to the specification, but is: git@github.com:siimon/prom-client.git`

![image](https://github.com/spdx/tools-python/assets/19174297/af592123-09b3-4cce-9957-dbee2ec1445a)

Could the specification be incorrect in this case and the tool should be able to adjust it?

maxhbr:
[Not sure what the specification says about that](https://spdx.github.io/spdx-spec/v2.3/package-information/#77-package-download-location-field). Download Location is either a URL or a "VCS location", where the "VCS location" is of the following form:
```
<vcs_tool>+<transport>://<host_name>[/<path_to_repository>][@<revision_tag_or_branch>][#<sub_path>]
```
Not sure if that example falls under URL or "VCS location", but the VCS Location seems to always contain `://`.

On the other hand, the specification has the example `git+git@git.myproject.org:MyProject` which is close to your form.

maxhbr:
Just checked https://validator.w3.org/checklink?uri=git%40github.com%3Asiimon%2Fprom-client.git&hide_type=all&depth=&check=Check and the download location you want to use is not a valid URL and fails with `error Error: 400 URL must be absolute`

PR #763, opened 2023-09-18
fix(validation): also allow URLs in download locations

### LINKED ISSUE #761: Unable to parse https://git.causal.agency/libretls/  as downloadLocation
Tool is failing to parse the downloadLocation in the following format:
`https://git.causal.agency/libretls/`

Following is the error:
`package download_location must be a valid download location according to the specification, but is: https://git.causal.agency/libretls/`

Is this expected to not support this URL format?

maxhbr:
All URLs are valid download locations.

This closes  #761 and #762.

### DIFF
--- a/src/spdx_tools/spdx/validation/uri_validators.py
+++ b/src/spdx_tools/spdx/validation/uri_validators.py
@@ -28,8 +28,8 @@ def validate_url(url: str) -> List[str]:
 
 
 def validate_download_location(location: str) -> List[str]:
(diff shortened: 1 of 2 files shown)


armintaenzertng (changes requested):
Please add the test cases from the issues

maxhbr:
I would prefer to not extend the current URL testcases, to avoid confusing URL semantics with recommendations on which URLs to use.

PR #768, opened 2023-10-05
Handle timezone-aware datetime objects when writing SPDX.

### LINKED ISSUE #766: spdx_tools created field does not matches the specification
spdx `created` [specification](https://spdx.github.io/spdx-spec/v2.3/document-creation-information/#69-created-field) expects 

> YYYY-MM-DDThh:mm:ssZ

as in the example:

> Created: 2010-01-29T18:30:22Z

However, the current implementation does not produce time in the expected format.

**Steps to reproduce**
1. Take attached example or [sbom from scratch](https://github.com/spdx/tools-python/blob/e08e4d25d721bb1ada1301d9d8e8515852387ea0/examples/spdx2_document_from_scratch.py) example and modify 

[Line 41](https://github.com/spdx/tools-python/blob/e08e4d25d721bb1ada1301d9d8e8515852387ea0/examples/spdx2_document_from_scratch.py#L41):     created=datetime.datetime.now(tz=datetime.timezone.utc),
[Line 75](https://github.com/spdx/tools-python/blob/e08e4d25d721bb1ada1301d9d8e8515852387ea0/examples/spdx2_document_from_scratch.py#L75): release_date=datetime.datetime.now(tz=datetime.timezone.utc),

2. Build SBOM
3. Resulting SBOM has 

>        "created": "2023-09-29T08:14:36+00:00Z",

This does not match the spec and is also warned by: https://tools.spdx.org/app/

[test_sbom.py.zip](https://github.com/spdx/tools-python/files/12762552/test_sbom.py.zip)

licquia:
Fixes #766.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ dependencies = ["click", "pyyaml", "xmltodict", "rdflib", "beartype", "uritools"
 dynamic = ["version"]
 
 [project.optional-dependencies]
-test = ["pytest", "pyshacl"]
+test = ["pytest", "pyshacl", "tzdata"]
 code_style = ["isort", "black", "flake8"]
 graph_generation = ["pygraphviz", "networkx"]
 development = ["black", "flake8", "isort", "networkx", "pytest"]
--- a/.github/workflows/install_and_test.yml
+++ b/.github/workflows/install_and_test.yml
@@ -33,6 +33,7 @@ jobs:
           python -m pip install --upgrade ./dist/*.whl
           python -m pip install pytest
           python -m pip install pyshacl
+          python -m pip install tzdata
           python -m pip install networkx
         shell: bash
       - name: Run tests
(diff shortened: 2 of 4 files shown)


licquia:
Looks like I didn't add the test dependency `tzdata` correctly, or missed something.  I'll look in the morning, or if someone wants to just ensure `tzdata` is installed as a test dependency everywhere it's needed, that should clear out the test failure.

pyproject.toml
```diff
@@ -28,7 +28,7 @@ dependencies = ["click", "pyyaml", "xmltodict", "rdflib", "beartype", "uritools"
 dynamic = ["version"]
 
 [project.optional-dependencies]
-test = ["pytest", "pyshacl"]
+test = ["pytest", "pyshacl", "tzdata"]
```
maxhbr, line 31: just a test dependency? I would have expected changes in non-test dependencies.

licquia: As I understand things, the library itself doesn't care about any timezone besides UTC (which is included in `datetime`).  I only need it for testing because I need to create a localized datetime object in the tests to check that we generate a correct ISO 8601 date string from it.

armintaenzertng (approved):
LGTM

PR #785, opened 2024-01-09
spdx3: element_writer: switch from tab characters to two spaces

stanislaw:
Hello everyone,

This is my first issue to this repository. I am new to SPDX and apologize for possible misunderstanding of the intentions behind the code I am modifying.

I am integrating the SPDX 3.0 export to the StrictDoc tool, see https://github.com/strictdoc-project/strictdoc/issues/1541.

The starting point for this issue is the fact that the `Creation Information` block is suddenly nested with a tab character when I try to populate an `spdx3.*.SpdxDocument` and write to a string buffer. My IDE is set up to ignore tab characters, so the output is displayed as if the indentation level was 1 character wide.

Bildschirmfoto 2024-01-09 um 21 07 31

In this MR, the creation info is unindented, and a test is added to demonstrate the behavior I would expect to see.

### DIFF
--- a/tests/spdx3/writer/tag_value/test_write_document.py
+++ b/tests/spdx3/writer/tag_value/test_write_document.py
@@ -1,0 +1,45 @@
+# SPDX-FileCopyrightText: 2024 spdx contributors
+#
+# SPDX-License-Identifier: Apache-2.0
+import io
+from datetime import datetime
+
+from semantic_version import Version
+
+from spdx_tools.spdx3.model import CreationInfo, ProfileIdentifierType, SpdxDocument
+from spdx_tools.spdx3.writer.console.spdx_document_writer import write_spdx_document
+
+
+def test_render_creation_info():
+    fake_datetime = datetime(year=2024, month=1, day=1)
+    spec_version = Version("3.0.0")
+    creation_info = CreationInfo(
+        spec_version=spec_version,
+        created=fake_datetime,
+        created_by=[],
+        profile=[ProfileIdentifierType.SOFTWARE],
+    )
+    spdx_document = SpdxDocument(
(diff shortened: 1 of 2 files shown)


tests/spdx3/writer/tag_value/test_write_document.py
```diff
@@ -0,0 +1,42 @@
+import io
```
maxhbr, line 1: This file is missing a license Header

maxhbr:
I think that the indentation was intentional. As sub-structures within an object are also come with indentation in other places. Maybe we could indent with spaces instead of `\t` to be more stable.

stanislaw:
> I think that the indentation was intentional. As sub-structures within an object are also come with indentation in other places. Maybe we could indent with spaces instead of `\t` to be more stable.

This makes sense. Definitely it would make sense to replace `\t` with `\n\n` for example.

An extra thought regarding the indentation of creation info: in YAML things are indented when there is a key that bundles together indented content. On the screenshot above, the role of a "key" is played by a comment `Creation Information` which to me looks confusing visually because it is just a comment and on actual key.

In YAML this would look like:

```yaml
creation_info:
  spec_version: ...
  ...
```

I am ready to both engage or disengage, so how could we proceed? Is it worth to use this MR for replacing `\t` with `\n\n` everywhere?

maxhbr:
we would appreciate if you would make the changes to spaces. But you could also just put it into a new issue, if you want.

stanislaw:
> we would appreciate if you would make the changes to spaces. But you could also just put it into a new issue, if you want.

Please check the change I've made.

maxhbr (approved):
LGTM

PR #786, opened 2024-01-09
spdx3: write_snippet: fix the presentation of line and byte ranges

stanislaw:
Without this change, the ranges are printed incorrectly, see the screenshot.

Bildschirmfoto 2024-01-09 um 21 36 32


tests/spdx3/writer/tag_value/test_write_snippet.py
```diff
@@ -0,0 +1,25 @@
+import io
```
maxhbr, line 1: This file is missing a license Header

maxhbr (approved):
LGTM

maxhbr:
Since it is a debug output, we decided to stay with the old representation, instead of pretty printing it.

Issue #790, opened 2024-01-25
Slow for SBOMs with a large number of files + relationships


jonjohnsonjr:
This check iterates over the list `existing_relationships` up to 2 times: https://github.com/spdx/tools-python/blob/8050fd9c41a92c75ec2ba9eb10ed9a919c375fa9/src/spdx_tools/spdx/parser/jsonlikedict/relationship_parser.py#L162-L172

And it's called for every file: https://github.com/spdx/tools-python/blob/8050fd9c41a92c75ec2ba9eb10ed9a919c375fa9/src/spdx_tools/spdx/parser/jsonlikedict/relationship_parser.py#L144-L157

So if `F` is the number of files and `R` is the number of relationships, we are doing `O(F * R)` comparisons of relationships (which seem not cheap individually).

And given that this seems to expect to find a relationship for every file, we know that `R is >= F`, so this is at least `O(R^2)`.

Looks quadratic to me.

With the caveat that I'm not a python expert, this being a list seems to be the issue: https://github.com/spdx/tools-python/blob/8050fd9c41a92c75ec2ba9eb10ed9a919c375fa9/src/spdx_tools/spdx/parser/jsonlikedict/relationship_parser.py#L55-L57

Is it possible to turn that into a set so these membership tests are O(1) instead of O(N)?

jonjohnsonjr:
As an example:

```
$ curl -L https://packages.wolfi.dev/os/aarch64/renovate-37.151.0-r0.apk | tar -Oxz var/lib/db/sbom/renovate-37.151.0-r0.spdx.json > sbom.spdx.json

$ wc -c sbom.spdx.json
 36350622 sbom.spdx.json

$ cat sbom.spdx.json | jq '.files[]' -c | wc -l
   30193

$ cat sbom.spdx.json | jq '.relationships[]' -c | wc -l
   30193

# 30193 * 30193 = 911617249, which is a pretty big number for python

$ time ntia-checker -v --file sbom.spdx.json

Is this SBOM NTIA minimum element conformant? True

Individual elements                            | Status
-------------------------------------------------------
All component names provided?                  | True
All component versions provided?               | True
All component identifiers provided?            | True
All component suppliers provided?              | True
SBOM author name provided?                     | True
SBOM creation timestamp provided?              | True
Dependency relationships provided?             | True

No components with missing information.

real	8m57.222s
user	8m57.118s
sys	0m0.085s
```

PR #792, opened 2024-01-29
Fix several accidentally quadratic functions

jonjohnsonjr:
Fixes https://github.com/spdx/tools-python/issues/790

I am by no means a python expert, please let me know if there are more idiomatic ways to solve this.

After this change:

```
$ time pyspdxtools -i sbom.spdx.json
pyspdxtools -i sbom.spdx.json  1.57s user 0.06s system 95% cpu 1.708 total
```

Note that this document previously took almost 9 minutes to validate, so this is ~300x faster for that particular document.


jspeed-meyers:
cc @meretp 

This patch grew out of the use of `ntia-conformance-checker`. Thanks, @meretp, for helping with the transition to the new `tools-python` version for `ntia-conformance-checker` a few months back.

src/spdx_tools/spdx/model/document.py
```diff
@@ -81,3 +81,6 @@ def __init__(
         relationships = [] if relationships is None else relationships
         extracted_licensing_info = [] if extracted_licensing_info is None else extracted_licensing_info
         check_types_and_set_values(self, locals())
+
+    def __hash__(self):
+        return id(self)
```
maxhbr, line 86: documents are not immutable and therefore hashing might run into inconsistencies. Most use-cases would probably expect a content dependent hash.

src/spdx_tools/spdx/model/relationship.py
```diff
@@ -73,3 +73,6 @@ def __init__(
         comment: Optional[str] = None,
     ):
         check_types_and_set_values(self, locals())
+
+    def __hash__(self):
+        return hash("{} -> {} ({})".format(self.spdx_element_id, str(self.related_spdx_element_id), str(self.relationship_type)))
```
maxhbr, line 78: I think our `Relationships` are mutable and that does not work well with hashing. This can cause inconsistencies when used in hashsets.

maxhbr (changes requested):
I *think* that hashing does not work well with mutable objects. It can cause inconsistencies and misses on lookup.

Also: not all properties are included in the hashing which causes false positive collisions.

src/spdx_tools/spdx/model/document.py
jonjohnsonjr: Ah that's a shame.

I wouldn't expect for the document to get mutated during validation. Would it make sense to have some frozen/immutable variants of document and relationship that could be passed around the validators?

Or do you have any ideas that would be a better approach to take?

paulgibert: Agree that a hash should only be implemented on immutable objects _when it uses mutable properties of the object_ - most "useful" hashes do. This hash is not based on a mutable property which would make it safe and consistent although not very useful (The default implementation of __hash__ is just id(self) >> 4). Also note that __eq__ should always be defined when __hash__ is [[see here](https://docs.python.org/3.5/reference/datamodel.html#object.__hash__)]. Why define __hash__ here?

src/spdx_tools/spdx/model/relationship.py
paulgibert: This class needs a definition of __eq__ as well.

jonjohnsonjr:
I've update the PR with changes from @paulgibert that seem better than what I came up with :)

jspeed-meyers:
@maxhbr or @meretp -- Any thoughts on this other approach?

maxhbr (commented):
the `astuple` approach looks promising.

src/spdx_tools/spdx/model/document.py
licquia: I've used `functools.lru_cache` in the past to speed up slow loops; it watches the parameters and replaces a function call with a table lookup when that makes sense.  The nice thing is that it can catch mutations and know to call the function again without calling it every single time.  Not sure if that would help in this case; I'll try to take a look this weekend.

src/spdx_tools/spdx/model/document.py
```diff
@@ -81,3 +81,18 @@ def __init__(
...
         extracted_licensing_info = [] if extracted_licensing_info is None else extracted_licensing_info
         check_types_and_set_values(self, locals())
+
+
+def document_cache(func):
+    cache = {}
+
+    def cached_function(document: Document):
+        key = id(document)
+        if key in cache.keys():
```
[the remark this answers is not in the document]
maxhbr: Or in other words: does the cache key `id(document)` change, if I mutate some stuff inside the document?

Maybe a less global caching would be more appropriate. So that its scope is controlled in a way so that no mutation can happen in between.

maxhbr:
Hey, maybe it would be worth to connect in a video call?

src/spdx_tools/spdx/model/document.py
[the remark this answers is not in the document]
paulgibert: I don't think so. From the [docs](https://docs.python.org/3/library/functions.html#id):

```
id(object)

    Return the “identity” of an object. This is an integer which is guaranteed to be unique and constant for this object during its lifetime.
Two objects with non-overlapping lifetimes may have the same [id()](https://docs.python.org/3/library/functions.html#id) value.
``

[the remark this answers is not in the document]
Comment (maxhbr) [reply src/spdx_tools/spdx/model/document.py:91]: I think that this is exactly the problem I am afraid of

```python
>>> a = {1}
>>> id(a)
139842939239360
>>> a.add(2)
>>> a
{1, 2}
>>> id(a)
139842939239360
```

The object has the same ID and it does not change even if its content is mutated. So it is maybe not a good cache key.

jonjohnsonjr:
How would you feel about something like this instead? https://github.com/spdx/tools-python/pull/800

I suspect just changing those function signatures is probably not what we want to do (I assume it's a breaking change), but what if we kept the same signatures and introduced a parallel set of functions that takes a set instead of a list (list variant can just call set variant).

PR #795, opened 2024-02-04
Speedup create_list_without_duplicates function

### LINKED ISSUE #794: `create_list_without_duplicates` Function Can be Sped Up By Using Set
The current implementation of `create_list_without_duplicates` includes a relatively expensive check of if an element is in a list.

https://github.com/spdx/tools-python/blob/8050fd9c41a92c75ec2ba9eb10ed9a919c375fa9/src/spdx_tools/spdx/document_utils.py#L51-L57

Using a set to keep track of unique elements will speed up this function.

I did a time comparison between this current function and a function that uses a set.

```python
from typing import List, Any
import timeit

# Original implementation
def create_list_without_duplicates_original(list_with_potential_duplicates: List[Any]) -> List[Any]:
    list_without_duplicates = []
    for element in list_with_potential_duplicates:
        if element not in list_without_duplicates:
            list_without_duplicates.append(element)
    return list_without_duplicates

# Optimized implementation
def create_list_without_duplicates_optimized(list_with_potential_duplicates: List[Any]) -> List[Any]:
    seen_elements = set()
    list_without_duplicates = []
    for element in list_with_potential_duplicates:
        if element not in seen_elements:
            seen_elements.add(element)
            list_without_duplicates.append(element)
    return list_without_duplicates

# Test data
test_data = list(range(1000)) * 10

# Measure time for the original implementation
original_time = timeit.timeit(lambda: create_list_without_duplicates_original(test_data), number=1000)

# Measure time for the optimized implementation
optimized_time = timeit.timeit(lambda: create_list_without_duplicates_optimized(test_data), number=1000)

print(f"Original implementation time: {original_time:.1f} seconds")
print(f"Optimized implementation time: {optimized_time:.1f} seconds")
```

The results are:

```bash
Original implementation time: 35.6seconds
Optimized implementation time: 0.2seconds
```

That original function is about 178x slower. I'll put in a PR for discussion.

jspeed-meyers:
Fix #794.

### DIFF
--- a/src/spdx_tools/spdx/document_utils.py
+++ b/src/spdx_tools/spdx/document_utils.py
@@ -49,9 +49,12 @@ def create_document_without_duplicates(document: Document) -> Document:
 
 
 def create_list_without_duplicates(list_with_potential_duplicates: List[Any]) -> List[Any]:
+    seen_elements = set()
     list_without_duplicates = []
+
     for element in list_with_potential_duplicates:
-        if element not in list_without_duplicates:
+        if element not in seen_elements:
+            seen_elements.add(element)
             list_without_duplicates.append(element)
 
     return list_without_duplicates


jspeed-meyers:
It looks like there is a failure of `TypeError: unhashable type: 'Actor'`. This seems similar to issue #792.

The tests passed locally on my machine: macOS Version 14.2.1 with Python 3.11.7 -- Are the tests in the CI and the local tests different in some way? I apologize. I wouldn't have submitted this PR if I knew my local tests weren't a good representation of the CI tests.

Any ideas?

src/spdx_tools/spdx/document_utils.py
```diff
@@ -49,9 +49,12 @@ def create_document_without_duplicates(document: Document) -> Document:
 
 
 def create_list_without_duplicates(list_with_potential_duplicates: List[Any]) -> List[Any]:
+    seen_elements = set()
     list_without_duplicates = []
+
     for element in list_with_potential_duplicates:
-        if element not in list_without_duplicates:
+        if element not in seen_elements:
+            seen_elements.add(element)
```
maxhbr, line 57: Instead of storing the element, you could store the element ID or similar to https://github.com/spdx/tools-python/pull/792 store the `astuple(element)`. This could solve the current issues in the CI

maxhbr (changes requested):

jspeed-meyers:
On second thought: I'm not sure this change is worth it. Sorry for the noise!

PR #807, opened 2024-06-29
Remove top level 'sbom' embedding

alper:
The Github spdx JSON file I got looked like this:

CleanShot 2024-06-29 at 23 18 17@2x

The nesting trips up the parser entirely but is easily fixed.

### DIFF
--- a/src/spdx_tools/spdx/parser/json/json_parser.py
+++ b/src/spdx_tools/spdx/parser/json/json_parser.py
@@ -13,4 +13,7 @@ def parse_from_file(file_name: str, encoding: str = "utf-8") -> Document:
     with open(file_name, encoding=encoding) as file:
         input_doc_as_dict: Dict = json.load(file)
 
+    if 'sbom' in input_doc_as_dict.keys():
+        input_doc_as_dict = input_doc_as_dict.get("sbom", {})
+
     return JsonLikeDictParser().parse(input_doc_as_dict)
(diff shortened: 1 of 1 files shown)


armintaenzertng:
Hi @alper, thanks for your contribution. Unfortunately, you forgot to sign-off your commit. Please remove your current commit and replace it with signed-off commit as described in point 5 of the [contribution guideline](https://github.com/spdx/tools-python/blob/main/CONTRIBUTING.md).

alper:
OK. That's done. Do I need to create a test case for this?

armintaenzertng:
Tests are always welcome! :)

Also note that the pipeline is currently broken, will be fixed in #814

armintaenzertng:
In addition to a test it might be good to not just silently parse non-conformant SBOMs such as these but alert the user that this was a faulty SBOM and that it has been parsed by discarding the top-level key.

maxhbr:
I just navigated to https://github.com/aboutcode-org/scancode-toolkit/network/dependencies and used the `Export SBOM` button and got a response containing the valid SPDX file without the surrounding `"sbom"` key. Are you using an API?

Thank you for your conrtibution, but I would close this for now. Please reopen if this is still valid.

alper:
Hi! Thanks for checking this.

You're right, if I write the sbom key to a file, I think it works:

```
  fs.writeFileSync(
    'sbom/choco-backend.json',
    JSON.stringify(response.data.sbom, null, 2)
  );
```

PR #816, opened 2024-08-02
[issue-788] fix tag-value parser to allow NONE and NOASSERTION for pa…

meretp:
…ckage source info as they are valid strings

fixes 788

### DIFF
--- a/src/spdx_tools/spdx/parser/tagvalue/parser.py
+++ b/src/spdx_tools/spdx/parser/tagvalue/parser.py
@@ -180,7 +180,7 @@ class Parser:
         "file_comment : FILE_COMMENT text_or_line\n "
         "file_license_concluded : FILE_LICENSE_CONCLUDED license_or_no_assertion_or_none\n "
         "package_name : PKG_NAME LINE\n description : PKG_DESCRIPTION text_or_line\n "
-        "summary : PKG_SUMMARY text_or_line\n source_info : PKG_SOURCE_INFO text_or_line\n "
+        "summary : PKG_SUMMARY text_or_line\n source_info : PKG_SOURCE_INFO text_or_line_including_no_assertion\n "
         "homepage : PKG_HOMEPAGE line_or_no_assertion_or_none\n "
         "download_location : PKG_DOWNLOAD_LOCATION line_or_no_assertion_or_none\n "
         "originator : PKG_ORIGINATOR actor_or_no_assertion\n supplier : PKG_SUPPLIER actor_or_no_assertion\n "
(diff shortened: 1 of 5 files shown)


meretp:
based this PR on #814 to (hopefully) have a successful CI

src/spdx_tools/spdx/parser/tagvalue/parser.py
```diff
@@ -180,7 +180,7 @@ def p_current_element_error(self, p):
         "file_comment : FILE_COMMENT text_or_line\n "
         "file_license_concluded : FILE_LICENSE_CONCLUDED license_or_no_assertion_or_none\n "
         "package_name : PKG_NAME LINE\n description : PKG_DESCRIPTION text_or_line\n "
-        "summary : PKG_SUMMARY text_or_line\n source_info : PKG_SOURCE_INFO text_or_line\n "
+        "summary : PKG_SUMMARY text_or_line\n source_info : PKG_SOURCE_INFO text_or_line_including_no_assertion\n "
```
armintaenzertng, line 183: The problem at hand can also occur for other `text_or_line` fields apart from `source_info`. Basically all properties that use `text_or_line`. So I would believe that the whole `text_or_line` property has to be adapted, not only for the source information case.

armintaenzertng: Also, the same problem will arise when somebody uses `NONE`. So best solve this one here, too.

tests/spdx/parser/tagvalue/test_package_parser.py
```diff
@@ -83,6 +83,26 @@ def test_parse_package():
     assert package.valid_until_date == datetime(2022, 1, 1, 12)
 
 
+def test_parse_package_with_no_assertion_as_source_info():
```
armintaenzertng, line 86: Ideally, this test should include all fields that can sport text only, all filled with either NONE or NOASSERTION. Might be bit overkill, though.
However, at least a field filled with NONE should be present.

armintaenzertng, line 86: We might want to rename this to `test_parse_none_and_no_assertion_as_text` or something. Otherwise we emphasize the source info despite it not being a special case.

meretp: I agree, I also moved the test to the `test_tag_valu_parser.py`module as it is not specific to packages.

meretp: Adding all cases would be overkill imho but I added a field that contains NONE and is parsed as text.

maxhbr:
@meretp  & @armintaenzertng  : can this be merged?

armintaenzertng (approved):

Issue #825, opened 2024-09-30
Validation for download_location does not respect URI with case insensitivity hostname and protocol


clabbenius:
To allow download location with case insensitive hostname and protocol the regular expression should be replaced in validation.

Suggestion:
(?i)(http:\/\/www\.|https:\/\/www\.|http:\/\/|https:\/\/|ssh:\/\/|git:\/\/|svn:\/\/|sftp:\/\/|ftp:\/\/)?([\w\-.!~*'()%;:&=+$,]+@)?[a-zA-Z0-9]+([\-.]{1}[a-zA-Z0-9]+){0,100}\.[a-zA-Z]{2,5}(:[0-9]{1,5})?(\/.*)?

armintaenzertng:
Fixed by #826

PR #826, opened 2024-09-30
downloadLocation URIs not case sensitive

clabbenius:
This is a fix for https://github.com/spdx/tools-python/issues/825


armintaenzertng (dismissed):
Thanks for the fix! I added some tests, I hope you don't mind.

armintaenzertng (approved):
Sorry, I messed up the DCO. Should be good now

Issue #827, opened 2024-10-01
PrimaryPackagePurpose field output OPERATING_SYSTEM instead of official value OPERATING-SYSTEM


loulou123546:
In SPDX v2.3, Package can have a property "PrimaryPackagePurpose" which support a defined set of value.

SPDX reference use "OPERATING-SYSTEM" value while the spdx_tools library output "OPERATING_SYSTEM" with underscore instead of dash. This may cause issue with ingestion of SPDX by other tools considered as invalid.

Reference: https://spdx.github.io/spdx-spec/v2.3/package-information/#724-primary-package-purpose-field

PR #828, opened 2024-10-01
[issue-827] fix mispelling of of OPERATING-SYSTEM

loulou123546:
A fix for issue #827

SPDX 2.3 reference use the term "OPERATING-SYSTEM" instead of "OPERATING_SYSTEM" which is was is output by this library.

Documentation: https://spdx.github.io/spdx-spec/v2.3/package-information/#724-primary-package-purpose-field


loulou123546:
Hello, can someone do a review of it?

armintaenzertng:
Hi @loulou123546, thanks for your contribution! This has always been ambiguous as the json schema did not match the specification you cite. However, by now, the json schema allows for both dashes and underscores, so we could indeed switch to the more spec-conform dashes.
[There is also](https://github.com/spdx/spdx-spec/pull/836) the `referenceCategory` which has the same problem ([see here in our code](https://github.com/spdx/tools-python/blob/a25937fa18ef3cfc57192502924a6ae723e7057d/src/spdx_tools/spdx/jsonschema/external_package_ref_converter.py#L22)). We should change that to use dashes there, too, so that we stay consistent, I think.
Do you want to take care of that, too?
I have to look into why the pipeline does not start for this PR...

PR #829, opened 2024-10-09
Update spdx3.model to SPDX v3.0.1

bact:
Update spdx_tools.spdx3.model to the latest in [SPDX v3.0.1](https://spdx.github.io/spdx-spec/v3.0.1/annexes/rdf-model/)

- Move Core Profile-related files from `spdx_tools.spdx3.model` to `spdx_tools.spdx3.model.core`
- Complete entries in vocabularies (enums) (e.g. HashAlgorithm, RelationshipType)
- Fix type in class initialization (List cannot be None)
- AIPackage: `sensitive_personal_information` -> `use_sensitive_personal_information`
- Build: `parameters` -> `parameter`
- Core: `imports` -> `import_` (cannot use "import", as it is Python's reserved)
- Dataset: `Dataset` -> `DatasetPackage`
- DatasetPackage: `sensitive_personal_information` -> `has_sensitive_personal_information`
- ExternalIdentifierType: `PURL` -> `PACKAGE_URL`
- RelationshipType: `concludedLicense` -> `hasConcludedLicense`
- RelationshipType: `declaredLicense` -> `hasDeclaredLicense` 

Mapping between spdx2 and spdx3 (that is required for conversion) is not yet in this PR.


goneall:
@maxhbr - Can you take a quick look and give @bact feedback

Discussed on the 15 Oct 2024 tech call

bact:
We might abandon this PR if we can generate the Python code from the model ontology file. I think that's probably a better approach in terms of maintenance.

keithofox:
I agree with you

alcroito:
Could you share any news when we might expect spdx / tools-python to support deserializing / validating v3.0.1 spdx documents?

Seems like this PR has been hanging for a while.

goneall:
@alcroito - we've been having some discussions on this topic on our weekly [SPDX Implementers meeting](https://github.com/spdx/meetings?tab=readme-ov-file#implementers-group-meetings).

We now have the generated model files that implement SPDX 3.0.1 available in a separate repo: https://github.com/spdx/spdx-python-model

The current thinking is we would refactor this tools-python code to use the generated model files, but that is quite a bit of work.

One possibility is to review and merge these changes now and make the design changes later.

@maxhbr @bact  - any thoughts on the approach?

bact:
I have take a look at [spdx-python-model](https://github.com/spdx/spdx-python-model). We can try refactor the code to use the generated model.

One of the things that the generated model is lacking compared to the spdx3.model in tools-python is the access to properties inside a class directly. You have to iterate the class members to get property names. Code may be more verbose.

What I have in mind:

1) Packaging, versioning, and making spdx-python-model available from PyPI (this is to ease the development/deployment in further steps)
2) Create (automatically?) a light layer that expose classes' properties
3) Refactor spdx3 code in tools-python

But yes that will be quite a work.

bact:
Maybe we can try also try to to rerun the CI test of this PR?

bact:
I will close this.
We may able to use https://github.com/spdx/spdx-python-model for SPDX 3 support.

With new shacl2code, the next release of spdx-python-model should provide more type annotations that will allow easier development.

PR #852, opened 2025-05-14
849-replace-underscore-with-dash-package-manager

mneumei:
Resolves the issue 849 to comply with the spdx.2.3 standard where PACKAGE_MANAGER is replaced by PACKAGE-MANAGER


mneumei:
Dear maintainers, please review

mneumei:
@meretp, i have a similar issue as @billie-alsup, where the reader will go strictly for the spec, and only allow for dashes. In #[793](https://github.com/spdx/spdx-spec/pull/793) there were also comments that the default should meet the spec. Would it be more helpful to make it selectable?

PR #853, opened 2025-06-30
Correct add swaps information to spdx 2 to 3

harrysarson:
When converting from `XXX_DEPENDENCY_OF` to `DEPENDS_ON` we need to swap
the dependency direction.


harrysarson:
Hey anything I can do to move this forward?

PR #895, opened 2026-06-17
Add the py.typed marker to the spdx3 package

Builder106:
## What
Adds a `py.typed` marker to the `spdx3` package (`src/spdx_tools/spdx3/py.typed`).

## Why
`spdx3` is typed inline like `spdx`, but it's missing the [PEP 561](https://peps.python.org/pep-0561/) `py.typed` marker. Without it, type-checkers (mypy, pyright) treat `spdx_tools.spdx3` as untyped and ignore its annotations for downstream consumers. `spdx` already ships `src/spdx_tools/spdx/py.typed`; this brings `spdx3` in line.

## Notes
No packaging change needed — the existing `[tool.setuptools.package-data]` rule `"*" = ["py.typed"]` already ships the marker from every package, so the new file is included automatically.

## Verification
Built the wheel (`python -m build`) and confirmed `spdx_tools/spdx3/py.typed` is now included alongside `spdx_tools/spdx/py.typed`.

Fixes #865


Builder106:
Ready for review.

Builder106:
Checking in on this since it's been open for a bit. Let me know if you need any changes made or if there's anything blocking review. @armintaenzertng @bact

PR #897, opened 2026-06-23
Read SPDX 3 by spdx-python-model

bact:
- Implement a thin layer for SPDX 3 (any 3.x version) support
- Using [`spdx-python-model`](https://github.com/spdx/spdx-python-model) as binding for SPDX 3 data model
- Use function names and structure that follow existing public SPDX 2 API of `tools-python`

This PR is trying to be small, focusing on ground work (format detection, etc), and demonstrate one core functionality (reading a file).

--

A `SpdxObjectSet` [Protocol](https://typing.python.org/en/latest/spec/protocol.html) structural type is introduced for convenience of type checking.
It will allow functions to have a type that is not specific to a SPDX 3.x minor version (`spdx_python_.model.v3_0_1.SHACLObjectSet` and `spdx_python_.model.v3_1.SHACLObjectSet` are considered different types).

The plan forward is to use the `spdx_tools.spdx3.SpdxObjectSet` (and `spdx_tools.spdx3.SpdxObject`, etc) protocol that derived directly from SPDX 3 RDF+SHACL (generated via `spdx-python-model`) as SPDX 3 data models' representation -- and remove the entire `src/spdx_tools/spdx3/model` to avoid maintenance of the multiple model versions (SPDX 3.0, SPDX 3.1, etc).


bact:
Note: Alternatively, we can implement the `SpdxObjectSet`[Protocol](https://typing.python.org/en/latest/spec/protocol.html) in `spdx-python-model` instead. (see full proposal at https://github.com/spdx/spdx-python-model/issues/45; and PRs at https://github.com/JPEWdev/shacl2code/pull/114 and https://github.com/spdx/spdx-python-model/pull/52)

This will allow anyone, including the more complete SPDX 3 implementation like ones in PR #898, to use the abstraction and freed them from managing model versions.

PR #900, opened 2026-07-09
Fix ExternalDocumentRef checksum parsing to accept all algorithms

Atishyy27:
Fixes #864

## Problem

The tag-value parser cannot parse an `ExternalDocumentRef` whose checksum uses any algorithm other than SHA1:

```
ExternalDocumentRef: DocumentRef-x https://example.com SHA256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
```

fails with `Couldn't match Checksum`, while the identical algorithm is accepted for `File` and `Package` checksums.

## Root cause

`ExternalDocumentRef` doesn't go through the lexer's `t_CHECKSUM` token (`lexer.py:140`), which already recognises every member of `ChecksumAlgorithm`. Instead it re-implements checksum matching with its own inline regex (`parser.py:288`):

```python
external_doc_ref_regex = re.compile(r"(.*)(\s*SHA1:\s*[a-f0-9]{40})")
```

That hardcodes both the algorithm (`SHA1`) and the digest length (`{40}`).

## Fix

Widen the regex to the same algorithm set the lexer already accepts.

## Behaviour change worth calling out

The new regex matches `[a-f0-9]*` rather than `[a-f0-9]{40}`, so **digest-length enforcement moves from parse time to validation time**.

This is deliberate, and it makes `ExternalDocumentRef` consistent with how `File` and `Package` checksums already behave: length is checked by `validation/checksum_validator.py`, which holds the `algorithm_length` table (`SHA1 -> 40`, `SHA256 -> 64`, …) and applies `^[0-9a-f]{N}$`. Keeping a length in the parser regex would have meant duplicating that table in a second place, and would still have been wrong for the 16 algorithms it didn't cover.

Concretely: `SHA1: afded` now *parses*, and is then rejected by the validator rather than by the parser. Because of that, one existing negative test in `test_creation_info_parser.py` was updated from an unparseable-checksum case to one that is still unparseable under the new regex:

```diff
- "ExternalDocumentRef: Document_ref document_uri SHA1: afded"
+ "ExternalDocumentRef: Document_ref document_uri MD7: afded"
```

`MD7` is not a real algorithm, so the case still exercises "parser rejects a malformed checksum". If reviewers would rather keep parse-time length checking, I'm happy to rework this — it just implies deriving the lengths from `checksum_validator.algorithm_length` instead of hardcoding them.

## Tests

Added `test_parse_external_document_ref_with_sha256_checksum`, using the SHA256 of the empty string (64 hex chars) so the parsed document also passes `validate_checksum`.

Verified locally on Windows (Python 3.10):

- new test fails on the parent commit, passes with the fix
- full suite: `1037 passed, 3 skipped`
- `isort`, `black`, `flake8` clean on both touched files


Atishyy27:
Hi @armintaenzertng, just a gentle bump on this when you have time.

This fixes #864 by allowing valid SHA256 ExternalDocumentRef checksums instead of only accepting SHA1. The PR is currently green and DCO is passing.

Happy to rebase or make any changes if needed. Thanks!

